PR #116, opened 2022-06-08
Do not open browser window

ghost:
Opening browser window does not work nice with multiple browsers. It picks one which is "default" which is not necessary the one which is being used at the moment :-)


matevz (approved):
I agree, this annoyed me too.

tjanez (approved):
+1 from me. I was annoyed at this too.

PR #117, opened 2022-06-09
Make circulation schedule chart interactive

ghost:
Make circulation chart interactive. It allows to see exact data points, zoom in-and-out etc.

| Before | After |
| --- | --- |
| before  |  after |


docusaurus.config.js
```diff
@@ -29,14 +29,27 @@ const config = {
...
-	    blog: false,
+        blog: false,
         theme: {
           customCss: require.resolve('./src/css/custom.css'),
         },
       }),
     ],
   ],
   plugins: [
+    /*function (context, options) {
```
tjanez: Why is this part commented out?

src/components/trs.csv
```diff
@@ -0,0 +1,6 @@
+﻿,0,3,6,9,12,15,18,21,24,27,30,33,36,39,42,45,48,51,54,57,60,63,66,69,72,75,78,81,84,87,90,93,96,99,102,105,108,111,114,117,120
```
tjanez: Is this redundant?

package.json
```diff
@@ -24,9 +24,11 @@
     "broken-link-checker": "^0.7.8",
     "clsx": "^1.1.1",
     "markdownlint-cli2": "^0.4.0",
+    "plotly.js": "^2.12.1",
```
lukaw3d: Hm, this is massive. It adds `Disk size with unique dependencies: "129.67MB"`

plotly.js-basic-dist would probably be equally good at `Disk size with unique dependencies: "2.72MB"`, and loaded with `createPlotComponent`

tjanez: Did you explore if that is possible?

lukaw3d: Nope

lukaw3d: But it has to be :D otherwise DELETE react-plotly.js

lukaw3d:
- Renamed everything
- Added fallback SVG for distribution chart
- Squashed
- Picked SVG rewards chart from https://github.com/oasisprotocol/docs/pull/158
- Picked CSV downloads from https://github.com/oasisprotocol/docs/pull/158

| Before | After |
| --- | --- |
| before  |  after |

kostko:
Hm the chart on the right doesn't seem entirely correct, e.g. for core contributors there should be a cliff at T+12 (value should be zero before) but here it looks like there is something between T+9 and T+12 which was not the case.

lukaw3d:
Hm, staking incentives SVG also looks different

tjanez:
> for core contributors there should be a cliff at T+12 (value should be zero before) but here it looks like there is something between T+9 and T+12 which was not the case.

Yeah, this is a consequence of how Plotly.js interpolates data points. By default, it uses linear interpolation.

Let me see if we can have something like in the old chart.

lukaw3d:
- Replaced monthly offsets with short date strings
- Improved tooltip
- Added dark theme
- Added empty months to reduce linear interpolation and match old chart https://github.com/oasisprotocol/docs/pull/117/commits/3ef7f725d6ca1270a112155103c9e00cb5e7cf45
- Reverted rewards chart changes
- Updated changelog

I'll squash now

kostko (approved):
Very nice!

tjanez (approved):
Thanks for this amazing work, @lukaw3d!

PR #120, opened 2022-06-10
ci: Add broken links check

matevz:


.github/workflows/blc.yml
```diff
@@ -0,0 +1,31 @@
+name: Broken link checker
```
tjanez: Can you rename the file to `ci-links.yml`?

tjanez: Also, can you add `name` fields to steps and comment them similar to https://github.com/oasisprotocol/oasis-core/blob/master/.github/workflows/ci-lint.yml?

tjanez (commented):
Can you group the changes into 3 commits:
- Add ci-links.yml
- Change config
- Fix actual broken links

tjanez:
Also, can you set it up so it is also run daily besides for PRs?

matevz:
I noticed the CI is occasionally getting error 429 from github because there are too many requests. Let's try to limit those.

matevz:
> I noticed the CI is occasionally getting error 429 from github because there are too many requests. Let's try to limit those.

Testing [linkinator](https://github.com/JustinBeckwith/linkinator) which supports multiple retries and multiple levels of verbosity.

matevz:
Tested linkinator and it has some nice advantages over broken-link-checker:
- retries on errors like 429 or if the site is not yet up (e.g. when running `yarn serve`) with increasing timeouts
- supports different verbosity levels (useful for reporting broken links only)
- supports config file (no need to pass everything as command line in `packages.json`)

As far as I tested it locally and with github actions, it is more robust than broken-link-checker and should be sufficient for the cron jobs.

tjanez (commented):
Can you split the last commit "ci: Add ci-links.yml" into 2 parts:
- Adding ci-lint.yml
- Changing broken links checker to linkinator

Otherwise, looks good!

tjanez (approved):

Issue #143, opened 2022-06-28
Reduce needlessly massive images


lukaw3d:
List of files >1MB:
- 1.23 MB https://github.com/oasisprotocol/docs/blob/main/docs/oasis-network-primer/images/Uni_Map_updated.png
- 1.47 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/switch_to_emerald.png
- 1.54 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/send_tokens_to_wormhole.png
- 1.58 MB https://github.com/oasisprotocol/docs/blob/main/docs/oasis-network-primer/images/Group_8.png
- 1.63 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/redeem_tokens.png
- 1.64 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/send_tokens.png
- 2.11 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/select_source_dest_amount.png
- 2.25 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/index.png
- 2.26 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/wallet/ledger/6b_ledger_oasis_ready.png
- 2.82 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/connect_wallet.png
- 2.89 MB https://github.com/oasisprotocol/docs/blob/main/docs/general/images/manage-tokens/wormhole/redeem_tokens2.png
- 3.63 MB https://github.com/oasisprotocol/docs/blob/main/docs/oasis-network-primer/images/ecosystem-network-01.jpeg

tjanez:
After #146 , 1 more image larger than 1 MB remains:
- 1.8 MB https://github.com/oasisprotocol/docs/blob/main/docs/oasis-network-primer/images/ecosystem-network-01.jpg

PR #146, opened 2022-06-30
Control sizes of large images

buberdds:
#143 

Options:
1) use official `plugin-ideal-image` to control large images
    - plugin is not bundled into `@docusaurus/preset-classic` preset like most of other official plugins
    - renaming to `.mdx` is not necessary, but might be good for third-party tools according to docs

2) manually replace images in repo and don't add `plugin-ideal-image`

3) Any other ideas?

* `6b_ledger_oasis_ready` does not need resizing. Switching to jpg is enough.


matevz:
Nice catch. Can you also reformat the rest of the images in this PR accordingly and/or find substitutes?

lukaw3d:
@matevz do you know if we have svg sources of e.g. https://github.com/oasisprotocol/docs/blob/main/docs/oasis-network-primer/images/Group_8.png

lukaw3d:
> option 2: manually replace images in repo and don't add plugin-ideal-image

I would prefer this, but sadly it won't reduce the repo size without rewriting history :/

lukaw3d:
Tried to optimize with https://squoosh.app/. docs/docs folder size: 43.9 MB -> 13.1 MB

Seems like jpgs lost some vibrancy.

buberdds:
Thanks for sharing, never heard about it before. Will explore options with resizing now.

lukaw3d (approved):

matevz:
Why did image brightness change? e.g. https://github.com/oasisprotocol/docs/pull/146/files#diff-c8ac17fbe04959322612fa9ece20f8c658da0ab93ce11bf35368b8d505a8ddb5

I think screenshots should be as they are. Lossless compression and resolution reduction from highdpi desktops are ok, but not changing colors/brightness.

matevz (approved):
Good work. Just check why the brightness of some images changed.

tjanez:
> but not changing colors/brightness.

I wouldn't worry about brightness too much currently.

Let's rather make sure we have a process that will prevent uploading PNG images that can't be compressed well.

tjanez (approved):
Thanks, @buberdds!

tjanez:
Follow up issues: #153 and #154.

lukaw3d:
Fixed colors by converting from "DELL U3818DW" color profile to "sRGB" and then removing it
`convert "$i" -profile /usr/share/color/icc/colord/sRGB.icc -strip`

tjanez:
> Fixed colors by converting from "DELL U3818DW" color profile to "sRGB" and then removing it
> `convert "$i" -profile /usr/share/color/icc/colord/sRGB.icc -strip`

Nice :slightly_smiling_face:

PR #147, opened 2022-06-30
Update common-staking-info.md

CrazySerGo:


docs/general/manage-tokens/advanced/oasis-cli-tools/common-staking-info.md
```diff
@@ -77,7 +77,7 @@ ParaTime, the entity for which the node is registered would need to satisfy the
 * Entity registration staking threshold (currently 100 tokens),
```
kostko, line 77: Can you also update the value a few lines above (line 66, I can't add a comment there) where it says (it is in base units):
```json
      "node-compute": "1000000000000000"
```

kostko (commented):
Nice catch, thanks!

kostko (approved):

PR #149, opened 2022-07-01
Added a simple systemd unit example.

valentinbud:
Signed-off-by: Valentin Bud <valentinbud@munay.network>


Yawning:
I would be in favor of adding something like this.

Yawning (approved):

PR #158, opened 2022-07-12
oasis-network-primer: Replace bitmap charts with SVG

matevz:
Simpler alternative to #117, if we decide not to bundle interactive charts.

Adds:
- SVG versions of the charts
- uses expected dates instead of T+x
- CSV download option
- ODS download option (also source for the SVG images versions above)
- synchronized final token redistribution and token distribution schedule chart colors

Preview: https://deploy-preview-158--trusting-archimedes-14c863.netlify.app/oasis-network-primer/token-metrics-and-distribution/


lukaw3d:
Staking rewards data chart doesn't match the old chart data

| Before png | After svg |
| --- | --- |
|   |   |

tjanez:
Thanks, @matevz, but let's go with the variant in #117 and work on making the staking rewards chart interactive in a subsequent PR.

PR #168, opened 2022-07-15
Polish token distribution chart (timezone bug, mobile view)

lukaw3d:
https://deploy-preview-168--trusting-archimedes-14c863.netlify.app/oasis-network-primer/token-metrics-and-distribution/#circulating-supply

| Before | After |
| --- | --- |
| before  |  after |

### DIFF
--- a/docs/oasis-network-primer/token_distribution/Chart.tsx
+++ b/docs/oasis-network-primer/token_distribution/Chart.tsx
@@ -13,9 +13,12 @@ const cumulativeSum = (values: number[]) => {
   return values.map(value => acc += value);
 };
 
+/**
+ * Offset from 2020-11-18
+ */
 const formatMonthOffsets = (monthOffsets: number[]) => {
   return monthOffsets.map(monthOffset => {
-    const date = new Date('2020-11-18');
+    const date = new Date(2020, 10 /* zero-based */, 18);
     date.setMonth(date.getMonth() + monthOffset);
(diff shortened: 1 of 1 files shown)


docs/oasis-network-primer/token_distribution/Chart.tsx
```diff
@@ -13,9 +13,12 @@ const cumulativeSum = (values: number[]) => {
   return values.map(value => acc += value);
 };
 
+/**
+ * Offset from 2020-11-18
+ */
 const formatMonthOffsets = (monthOffsets: number[]) => {
   return monthOffsets.map(monthOffset => {
-    const date = new Date('2020-11-18');
+    const date = new Date(2020, 10 /* zero-based */, 18);
```
pro-wh, line 21: :information_source: monthIndex bug, not timezone bug

pro-wh (approved):
the date thing fix looks good. other changes look pretty undisruptive

docs/oasis-network-primer/token_distribution/Chart.tsx
lukaw3d: `Date('2020-11-18')` was midnight in UTC, so toLocaleDateString in some timezones returned 2020-11-17.
`Date(2020, 10 /* zero-based */, 18)` is midnight in local timezone, so toLocaleDateString should always return 2020-11-18.

I'd prefer a more intuitive `Date('2020-11-18').toUTCString` but it doesn't have formatting options :cry:

PR #198, opened 2022-08-11
docs: Fix discord invite link

matevz:
Fix expired discord invite link.


kostko (approved):
Does this need to be refreshed periodically?

matevz:
> Does this need to be refreshed periodically?

It shouldn't. Maybe the invite link was created by a user which left the server. :shrug:

PR #223, opened 2022-09-12
docs: Rename dapp -> dApp for consistency

matevz:
After thorough discussion, we decided to use *dApp* term for denoting a distributed app. This PR replaces all *dapp* occurrences with *dApp*.


kostko (approved):

matevz:
After reading [Ethereum docs on dapps](https://ethereum.org/en/dapps/), it seems the existing term *dapp* is more fit for abbreviating "distributed applications"?

Issue #225, opened 2022-09-12
Code snippets Remark plugin


matevz:
To render code snippets from examples in our docs, we should implement a Remark plugin which loads the example code from file and transforms specific markdown pattern (see below) to the standard code block markdown showing the example.

There is no standard markdown syntax for importing code snippets. A nice option would be to take the image loading syntax (`![alt text](filename.png "title")`) combined with the [markdown vuepress v2 syntax for importing code](https://vuepress.github.io/guide/markdown.html#import-code-blocks) and extend it with anchor support:
  ```
  ![code go](../../examples/path/to/my/file.go#region-or-line-number-in-github-syntax "someTitle.go")
  ```

`!` and `code` are literals. The second alt argument is language name and is optional. The link always contains relative path to the example file. Optionally, you can put anchor sign `#` after the filename followed by the region name inside the example file or the line number in github format, e.g. `L10-L13`. After the file path an optional title is provided (shown at the top of the code block).

If the markdown parser doesn't support this fancy syntax (e.g. github), it will still produce a readable `code go` link to the corresponding source file.

Regions of the snippets are defined in example files with `#region <region-name>` and `#endregion <region-name>` in the comments (taken from [vs code](https://code.visualstudio.com/docs/editor/codebasics#_folding) and used by the [vuepress v1 code snippet import plugin](https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets)).

Examples should be addressed using the relative path from the markdown file to example file. If the markdown file is a symbolic link, first obtain the *real* path and then look for the example file.

Using the code above, the remark plugin should generate the following markdown:
```
  ```go title="someTitle.go" href="https://github.com/oasisprotocol/oasis-sdk/examples/path/to/my/file.go#L13-L17"
     actual go file content
  ``` `
```

If the code title is set, then code block title is shown. The title is clickable and should take you to the github page showing the specified file with the corresponding highlighted lines based on the anchor `#`, if provided.

matevz:
Implemented in #236.

PR #236, opened 2022-09-19
Add support for importing code snippets from files

matevz:
Implements #225.

- Also adds jest to devDependency and basic tests for the plugin to CI.

## Note
URL link to the source file on github is there, but is currently not rendered in the html. This would probably need swizzling of the code block component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "deploy": "docusaurus deploy",
     "clear": "docusaurus clear",
     "serve": "docusaurus serve",
+    "test": "jest src",
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "blc": "linkinator http://localhost:3000/general",
@@ -45,6 +46,7 @@
   },
   "devDependencies": {
     "@types/plotly.js-basic-dist": "^1.54.0",
-    "@types/react-plotly.js": "^2.5.1"
+    "@types/react-plotly.js": "^2.5.1",
+    "jest": "^29.0.3"
   }
 }
--- a/src/editUrl.js
+++ b/src/editUrl.js
@@ -3,10 +3,10 @@ const fs = require('fs-extra');
 const path = require('path');
 
 const gitModules = {
-    'external/adrs/': 'https://github.com/oasisprotocol/adrs/edit/main',
-    'external/oasis-core/': 'https://github.com/oasisprotocol/oasis-core/edit/master',
-    'external/oasis-core-ledger/': 'https://github.com/oasisprotocol/oasis-core-ledger/edit/master',
-    'external/oasis-sdk/': 'https://github.com/oasisprotocol/oasis-sdk/edit/main',
+    'external/adrs/': 'https://github.com/oasisprotocol/adrs/{mode}/main/',
+    'external/oasis-core/': 'https://github.com/oasisprotocol/oasis-core/{mode}/master/',
+    'external/oasis-core-ledger/': 'https://github.com/oasisprotocol/oasis-core-ledger/{mode}/master/',
+    'external/oasis-sdk/': 'https://github.com/oasisprotocol/oasis-sdk/{mode}/main/',
 };
 
 /**
@@ -15,20 +15,32 @@ const gitModules = {
  * @type {import('@docusaurus/plugin-content-docs').EditUrlFunction}
  */
 function editLinkUrl(params) {
-    // Resolve symlink, if needed.
     const relFilepath = path.join(params.versionDocsDirPath, params.docPath);
-    const p = fs.realpathSync(relFilepath);
+    return linkUrl(relFilepath, 'edit');
+}
 
+/**
+ * Generates cross-repo view URL link (e.g. for viewing example sources).
+ */
+function viewLinkUrl(relFilePath) {
+    return linkUrl(relFilePath, 'blob');
+}
+
+function linkUrl(filename, mode) {
+    // Resolve composed paths and symlinks, if needed.
+    filename = fs.realpathSync(filename);
+
+    // Obtain relative filename to project root.
+    filename = path.relative(process.cwd(), filename)
     for (const r in gitModules) {
-        const regex = new RegExp( `.*${r}(.*)`);
-        if (regex.test(p)) {
+        if (filename.startsWith(r)) {
             // Extract relative filename inside the git submodule.
-            const file = regex.exec(p)[1];
-            return `${gitModules[r]}/${file}`;
+            return filename.replace(r, gitModules[r]).replace("{mode}", mode);
         }
     }
 
-    return `https://github.com/oasisprotocol/docs/edit/main/${relFilepath}`;
+    return `https://github.com/oasisprotocol/docs/{mode}/main/${filename}`.replace("{mode}", mode);
+
 }
 
-module.exports = editLinkUrl;
+module.exports = {editLinkUrl, viewLinkUrl};
--- a/src/remark/code-block-snippets.js
+++ b/src/remark/code-block-snippets.js
@@ -1,0 +1,105 @@
+// @ts-check
+const fs = require('fs-extra');
+const path = require("path");
+const visit = require('unist-util-visit');
+const viewLinkUrl = require('../editUrl.js').viewLinkUrl;
+
+// File name of the markdown file which references external code snippet.
+let mdFilePath = "";
+
+/**
+ * Searches for ![code](source-file) pattern and replaces it with the code block containing the
+ * content of the source-file. It also supports importing specific lines of code or regions
+ * inside the source file defined by #region and #endregion literals.
+ *
+ * Examples:
+ * ![code](../examples/my-example/shell-dump.txt)
+ * ![code go](../examples/my-example/src/main.go)
+ * ![code go](../examples/my-example/src/server.go "src/server.go")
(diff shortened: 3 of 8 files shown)


src/editUrl.js
```diff
@@ -15,20 +15,34 @@ const gitModules = {
...
     for (const r in gitModules) {
-        const regex = new RegExp( `.*${r}(.*)`);
-        if (regex.test(p)) {
+        const regex = new RegExp( `${r}(.*)`);
+        if (regex.test(filename)) {
             // Extract relative filename inside the git submodule.
-            const file = regex.exec(p)[1];
-            return `${gitModules[r]}/${file}`;
+            const file = regex.exec(filename)[1];
+            return `${gitModules[r]}/${file}`.replace("{mode}", mode);
```
lukaw3d: I'd simplify to
```js
    filename = path.relative(process.cwd(), filename)
    for (const r in gitModules) {
        if (filename.startsWith(r)) {
            // Extract relative filename inside the git submodule.
            const file = path.relative(r, filename)
            return `${gitModules[r]}/${file}`.replace("{mode}", mode);
```

or end with
`return filename.replace(r, gitModules[r]).replace("{mode}", mode)`
and add trailing slashes to gitModules

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+            node.value = code;
+        }
+    }
+
+    function transform(tree) {
+        visit(tree, ['image'], visitor);
+    }
+
+    return (root, vfile) => {
+        filePath = vfile.history[0]
```
lukaw3d: ew shared variable, but i guess it work

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+                // From-to range of line numbers.
+                let range = [];
+                // XXX: Online reading line-by-line from buffer causes some async JS abnormalities.
+                const codeSplit = code.split("\n");
+                code = "";
+
+                const linesRegex = /L(\d+)(-L(\d+))?/;
+                if (linesRegex.test(anchor)) {
+                    // Anchor contains a single line number (e.g. L14) or range (e.g. L10-L23).
+                    linesParts = linesRegex.exec(anchor);
```
lukaw3d: (This is a global var)

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+                    }
+
+                    // Sanity check.
+                    if (range.length != 2) {
+                        throw new Error(`region '${urlAnchor[1]}' not found in ${sourceFilePath}`);
+                    }
+                }
+
+                // Append line numbers to source URL even if region name was initially provided.
+                node.url += `#L${range[0]}`
```
lukaw3d: How come you're specifying URL on a type=code node?

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+
+            let code = fs.readFileSync(sourceFilePath, 'utf-8');
+            node.url = viewLinkUrl(sourceFilePath);
+
+            // Check, if anchor matches line number or region name inside the file.
+            const anchor = urlAnchor[1];
+            if (anchor) {
+                // From-to range of line numbers.
+                let range = [];
+                // XXX: Online reading line-by-line from buffer causes some async JS abnormalities.
```
lukaw3d: (I don't understand this comment at all :D )

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+            // Check for language as the second argument. e.g. ![code rust](some-file.rs)
+            node.lang = node.alt.split(" ")[1];
+
+            // Use image title for the code block title.
+            if (node.title) {
+                node.meta = `title="${node.title}"`;
+            }
+
+            // Source filename can have anchor symbol # followed by the line numbers or region name.
+            const urlAnchor = node.url.split("#");
```
lukaw3d: I think there's too many similarly named variables, so:
```suggestion
            const [relativeSourceFilePath, anchor] = node.url.split("#");
```
I'd also rename "anchor" to fragment - https://developer.mozilla.org/en-US/docs/Web/API/URL/hash

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+     * Examples:
+     * ![code](../examples/my-example/shell-dump.txt)
+     * ![code go](../examples/my-example/src/main.go)
+     * ![code go](../examples/my-example/src/server.go "src/server.go")
+     * ![code go](../examples/my-example/src/server.go#L3)
+     * ![code go](../examples/my-example/src/server.go#L5-L11 "src/server.go")
+     * ![code go](../examples/my-example/src/account.go#native-account "Structure of the Native account")
+     */
+    function visitor(node) {
+        if (node.alt && node.alt.startsWith("code")) {
```
lukaw3d:
```js
        if (!node.alt?.startsWith("code")) return;
```
to reduce one indentation

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * Examples:
+     * ![code](../examples/my-example/shell-dump.txt)
+     * ![code go](../examples/my-example/src/main.go)
+     * ![code go](../examples/my-example/src/server.go "src/server.go")
+     * ![code go](../examples/my-example/src/server.go#L3)
+     * ![code go](../examples/my-example/src/server.go#L5-L11 "src/server.go")
+     * ![code go](../examples/my-example/src/account.go#native-account "Structure of the Native account")
+     */
+    function visitor(node) {
```
lukaw3d: We can have some intellisense and type warnings with

```suggestion
     *
     * @typedef {import('mdast').Image} ImageNode
     * @typedef {import('mdast').Code} CodeNode
     * @param {Readonly<Omit<ImageNode, 'type'>> & CodeNode} node
     */
    function visitor(node) {
```

src/remark/code-block-snippets.js
```diff
@@ -0,0 +1,125 @@
...
+                node.url += `#L${range[0]}`
+                if (range[1]) {
+                    node.url += `-L${range[1]}`;
+                }
+            }
+
+            // Remove trailing newline, if any.
+            code = code.replace(/\n$/, "")
+            node.value = code;
+        }
```
lukaw3d: I'd refactor to:
```js
        const code = fs.readFileSync(sourceFilePath, 'utf-8');
        node.url = viewLinkUrl(sourceFilePath);

        // Remove trailing newline, if any.
        let snippet = code.replace(/\n$/, "")

        // Check, if anchor matches line number or region name inside the file.
        if (anchor) {
            // From-to range of line numbers.
            let range;
            const linesParts = /L(\d+)(-L(\d+))?/.exec(anchor)
            if (linesParts) {
                // Anchor contains a single line number (e.g. L14) or range (e.g. L10-L23).
                range = [
                    parseInt(linesParts[1]),
                    parseInt(linesParts[3] ?? linesParts[1]),
                ]
            } else {
                // Anchor references region with #region and #endregion literals.

                // +1/-1 to exclude lines with #region, and +1 for 1-based numbering
                range = [
                    code.split('\n').findIndex((line) => line.includes(`#region ${anchor}`)) + 1 + 1,
                    code.split('\n').findIndex((line) => line.includes(`#endregion ${anchor}`)) - 1 + 1,
                ]
                if (range[0] < 2) throw new Error(`"#region ${anchor}" not found`)
                if (range[1] < 0) throw new Error(`"#endregion ${anchor}" not found`)
                if (range[0] > range[1]) throw new Error(`"#region ${anchor}" appears after "#endregion ${anchor}"`)
            }

            // Append line numbers to source URL even if region name was initially provided.
            node.url += `#L${range[0]}-L${range[1]}`;

            // -1 for 1-based numbering
            snippet = code.split('\n').slice(range[0] - 1, range[1] - 1 + 1).join('\n')
        }

        node.value = snippet;
```

src/remark/code-block-snippets.js
matevz: I'd like to have clickable code title which brings you to the source file on github (with highlighted lines).

src/remark/code-block-snippets.js
matevz: Yeah, I haven't find a way to pass this option to specific plugin instance. This would be problematic, if compilation is ran with concurrency (which is not currently ;) ).

src/remark/code-block-snippets.js
matevz: I first implemented https://nodejs.org/api/readline.html, but it didn't work - the read content was always empty although the reading logic worked. Then I gave up and simply read complete file to memory and then split it to separate lines.

src/remark/code-block-snippets.js
matevz: Thanks! I'm not fond of re-reading the file twice, when region names provided, but it won't give big of an impact for real world examples since they are small.

lukaw3d (approved):

package.json
```diff
@@ -23,6 +24,7 @@
     "@easyops-cn/docusaurus-search-local": "^0.26.0",
     "@mdx-js/react": "^1.6.22",
     "clsx": "^1.1.1",
+    "jest": "^29.0.3",
     "linkinator": "^3.1.0",
     "markdownlint-cli2": "^0.4.0",
```
lukaw3d: (these would more accurately be devDependencies)

PR #245, opened 2022-09-27
code-block-snippets: Move code block in place of paragraph

matevz:
Followup to #236 .

This PR fixes the level of the code block node (from `root -> paragraph -> code block` to `root -> code block`). Wrong code block level resulted in admonition anomalies in subsequent paragraphs.

### DIFF
--- a/src/remark/code-block-snippets.js
+++ b/src/remark/code-block-snippets.js
@@ -8,9 +8,9 @@ const viewLinkUrl = require('../editUrl.js').viewLinkUrl;
 let mdFilePath = "";
 
 /**
- * Searches for ![code](source-file) pattern and replaces it with the code block containing the
- * content of the source-file. It also supports importing specific lines of code or regions
- * inside the source file defined by #region and #endregion literals.
+ * This plugin searches for ![code](source-file) pattern and replaces it with the code block
+ * containing the content of the source-file. It also supports importing specific lines of code or
+ * regions inside the source file defined by #region and #endregion literals.
  *
  * Examples:
  * ![code](../examples/my-example/shell-dump.txt)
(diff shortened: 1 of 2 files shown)


src/remark/code-block-snippets.js
```diff
@@ -24,8 +24,16 @@ let mdFilePath = "";
...
+
+/**
+ * Find paragraph -> image pattern and perform the replacement.
+ */
 function visitor(node) {
-    if (!node.alt?.startsWith("code")) return;
+    if (!node.children || node.children[0]?.type != 'image' || !node.children[0]?.alt?.startsWith("code")) return;
+
+    // Move code block in place of the paragraph node.
+    Object.assign(node, node.children[0]);
```
lukaw3d: I don't think this assignment is needed. Image node has properties:
```ts
type: 'image';
url: string;
alt?: string;
title?: string;
data?: TData;
position?: Position;
```

and below you set:
```ts
type: 'code';
value: string;
url: string;
lang?: string;
meta?: string;
```

so only `alt, title, data, position` would remain, but don't seem to change anything.

So a more readable alternative would be
```ts
    const imgNode = node.children[0];
    delete node.children;
    node.type = 'code';
    node.lang = imgNode.alt.split(" ")[1];
    ...
```

src/remark/code-block-snippets.js
```diff
@@ -24,8 +24,16 @@ let mdFilePath = "";
  * @typedef {import('mdast').Code} CodeNode
  * @param {Readonly<Omit<ImageNode, 'type'>> & CodeNode} node
  */
+
+/**
+ * Find paragraph -> image pattern and perform the replacement.
+ */
 function visitor(node) {
```
lukaw3d: New jsdoc breaks the param type declaration above.

Move most of the first jsdoc to plugin function
```js
/**
 * This plugin searches for ![code](source-file) pattern and replaces it with the code block
 * containing the content of the source-file. It also supports importing specific lines of code or
 * regions inside the source file defined by #region and #endregion literals.
 *
 * Examples:
 * ![code](../examples/my-example/shell-dump.txt)
 * ![code go](../examples/my-example/src/main.go)
 * ![code go](../examples/my-example/src/server.go "src/server.go")
 * ![code go](../examples/my-example/src/server.go#L3)
 * ![code go](../examples/my-example/src/server.go#L5-L11 "src/server.go")
 * ![code go](../examples/my-example/src/account.go#native-account "Structure of the Native account")
 */
function plugin(options) {
```

and move param type into visitor and update it
```js
/**
 * Find paragraph -> image pattern and perform the replacement.
 *
 * @param {Omit<import('mdast').Paragraph, 'type'> & import('mdast').Code} node
 */
function visitor(node) {
```

src/remark/code-block-snippets.js
```diff
@@ -24,8 +24,16 @@ let mdFilePath = "";
...
-    if (!node.alt?.startsWith("code")) return;
+    if (!node.children || node.children[0]?.type != 'image' || !node.children[0]?.alt?.startsWith("code")) return;
+
+    // Move code block in place of the paragraph node.
+    Object.assign(node, node.children[0]);
+    delete node.children;
```
lukaw3d, line 44: I'm not sure what issue this PR is supposed to solve, but this can eat up content:
```md
# Getting Started

![code js](../src/remark/code-block-snippets.test.js#test-region)
hello

world

## Use Oasis
```

| Before | After |
| --- | --- |
| Screenshot from 2022-09-27 21-37-56  | Screenshot from 2022-09-27 21-34-57  |

matevz: Good catch. I made a fix to require only a single child containing the image node. Otherwise it ignores the external code snippet.

lukaw3d: Perhaps it would be even better to throw an error in that case so developers notice

lukaw3d (approved):

src/remark/code-block-snippets.js
lukaw3d: Ah, it does give a warning
```
[WARNING] The image at "..../src/remark/code-block-snippets.test.js" can't be read correctly. Please ensure it's a valid image.
unsupported file type: undefined (file: ..../src/remark/code-block-snippets.test.js)
```

PR #274, opened 2022-10-17
docs/node: Bump Oasis Core to 22.2 for Testnet

kostko:


pro-wh (approved):
discussed on slack, we're upgrading testnet now

PR #283, opened 2022-10-28
Add signature precompile descriptions

jberci:
Document precompiles added in oasisprotocol/oasis-sdk#1176.

### DIFF
--- a/docs/dapp/sapphire/precompiles.md
+++ b/docs/dapp/sapphire/precompiles.md
@@ -4,39 +4,75 @@ description: Additional Sapphire precompiles for encryption and confidentiality
 
 # Precompiles
 
-In addition to the standard EVM precompiles, Sapphire provides three
-further cryptography-related ones to make some operations easier and
-cheaper to perform: x25519 key derivation, and Deoxys-II-based encryption
-and decryption.
+In addition to the standard EVM precompiles, Sapphire provides a number
+of further cryptography-related ones to make some operations easier and
+cheaper to perform: x25519 key derivation, Deoxys-II-based encryption
+and decryption, signing key generation, message digest signing and
+verification.
 
 These can be called in the same way as other precompiles by dispatching
 calls to specific well-known contract addresses, as described below.
-Input parameters should be packed into a contiguous memory region, with
-each chunk of data padded to 32 bytes as usual.
+
+Input parameters should be packed into a contiguous memory region with
+each chunk of data padded to 32 bytes as usual. The recommended way to
+construct parameter byte sequences in Solidity is with `abi.encode` and
+`abi.decode`, which will transparently handle things like putting
+`bytes` lengths in the correct position.
+
+## Library
+
+While it is possible to call the precompiles directly using Yul or, for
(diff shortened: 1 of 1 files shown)


kostko:
@jberci could you update this to reflect the new precompile arguments (also for other methods)? E.g. to be consistent with the [wrapper library](https://github.com/oasisprotocol/sapphire-paratime/blob/main/contracts/contracts/Sapphire.sol).

docs/dapp/sapphire/precompiles.md
```diff
@@ -4,22 +4,61 @@ description: Additional Sapphire precompiles for encryption and confidentiality
...
 Input parameters should be packed into a contiguous memory region, with
-each chunk of data padded to 32 bytes as usual.
+each chunk of data padded to 32 bytes as usual. Some precompiles have
+complex input parameter layouts, so it is recommended to use
+`abi.encode` and `abi.decode` to handle them; the signatures below are
+given with this in mind.
+
+## Library
+
+A few of the examples below show how to call the precompiles directly.
```
kostko: I think we should just make all examples use the wrappers as otherwise people will just copy these.

docs/dapp/sapphire/precompiles.md
```diff
@@ -45,68 +81,118 @@ function deriveSymmetricKey(bytes32 keyPublic, bytes32 keyPrivate) private view
...
+```solidity
+bytes32 key = ... ;
+bytes32 nonce = ... ;
+bytes memory text = "plain text";
+bytes memory ad = "additional data";
+bytes memory encrypted = Sapphire.encrypt(key, nonce, text, ad);
+bytes memory decrypted = Sapphire.decrypt(key, nonce, encrypted, ad);
+```
+
+## Public/Private Key Pair Generation
```
kostko:
```suggestion
## Public/Private Signing Key Pair Generation
```

docs/dapp/sapphire/precompiles.md
```diff
@@ -45,68 +81,118 @@ function deriveSymmetricKey(bytes32 keyPublic, bytes32 keyPrivate) private view
...
-        mstore(out, returndatasize())
-        returndatacopy(add(out, 0x20), 0, returndatasize())
-        mstore(0x40, add(add(out, 0x20), returndatasize()))
-        result := out
-    }
-}
+bytes memory seed = hex"0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef";
+bytes memory publicKey;
+bytes memory privateKey;
+(publicKey, privateKey) = Sapphire.generateKeyPair(Sapphire.SigningAlg.Ed25519Pure, seed);
```
kostko:
```suggestion
(publicKey, privateKey) = Sapphire.generateSigningKeyPair(Sapphire.SigningAlg.Ed25519Pure, seed);
```

kostko (approved):
Looks good, just some tweaks due to oasisprotocol/sapphire-paratime#116 being merged. The new precompile will also need to be documented but it is not yet in a release so no need to do it now.

Issue #308, opened 2022-11-17
Adjust diagrams for dark theme 


matevz:
Go through all diagrams and check that they are readable in the dark theme:
- [x] for SVG images drawn with drawio.com, reexport them with "Automatic" Appearance option and "Transparent background"
- [ ] for SVG images generated by mermaidjs, write a theme-sensitive CSS
- [x] get rid of the [enforced white background for all images](https://github.com/oasisprotocol/docs/blob/main/src/css/custom.css#L145-L148)

Transparent PNGs (do we have any?) should either be redrawn/reexported in SVG or designed in such a way that the text is readable regardless of the background. Non-transparent PNGs and JPEGs could maybe use [Docuaurus' themed images](https://docusaurus.io/docs/markdown-features/assets#themed-images) in the future?

matevz:
Possible solution would be the following:
- for any image (png, svg, jpeg) use the default `myimage.png` filename for light themes. For dark themes, check, if `myimage_dark.png` exists and use it instead. Otherwise keep the original image. So something like the [ThemedImage](https://docusaurus.io/docs/markdown-features/assets#themed-images), but automated.
- for .svg files, instead of having another image, cool kids now detect the theme and change coloring with inline CSS. For example check out the [current favicon](https://raw.githubusercontent.com/oasisprotocol/docs/b4b701b466b310eb4fbb9a895c978c0db8cbf608/static/img/favicon.svg) or the [ROFL diagram](https://raw.githubusercontent.com/oasisprotocol/oasis-sdk/50bab3f5ab606b3d2a737579200cc3f720d83409/docs/rofl/rofl.svg).

matevz:
drawio.com which we use for most of the hand-drawn diagrams now has theming support for SVG export. Selecting the "Automatic" theme option in the export dialog adds theme-sensitive CSS snippet in the SVG. This, and enabling transparent background gives best results in our docs.
![image](https://github.com/user-attachments/assets/b2986b7b-d3d2-498f-a39d-550a6a7a5817)

Final result:
| Light | Dark |
|---------|---------|
| ![image](https://github.com/user-attachments/assets/b878d353-eb46-4134-b088-cd92475ecb6d) | ![image](https://github.com/user-attachments/assets/1531a474-3c70-4a91-8ec3-a18d5f49017c) |

aefhm:
## TODO

[...]

PR #309, opened 2022-11-17
CSS: Add bright background to images when using dark theme

matevz:
A temporary hack to mitigate #308. Adds white background to images when switching to dark theme. 

| before | after |
|------------|--------|
| image | image |
| image |  image |

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -89,3 +89,8 @@ html[data-theme='dark'] .docusaurus-highlight-code-line {
 .theme-doc-sidebar-menu {
   font-size: 14px;
 }
+
+/* Hack to make .svg and .png diagrams readable in dark theme */
+html[data-theme='dark'] main img {
+  background-color: #fff;
+}


kostko (approved):

src/css/custom.css
```diff
@@ -89,3 +89,8 @@ html[data-theme='dark'] .docusaurus-highlight-code-line {
 .theme-doc-sidebar-menu {
   font-size: 14px;
 }
+
+/* Hack to make .svg and .png diagrams readable in dark theme */
+html[data-theme='dark'] main img {
```
lukaw3d, line 94: If we'll be adding some transparent themed images: `img:not([class^="themedImage"])`

(not the most stable element selector but, https://docusaurus.io/docs/markdown-features/assets#themed-images
compiles un-themed images as ``
and themed images as `` )

lukaw3d (approved):

src/css/custom.css
lukaw3d: oh. and `:not([src$='#gh-dark-mode-only'])`

matevz: Not really - I imagined when #308 is done properly, then each image must either be universal (good to read both on light and dark background) or themed (one image for light, other for dark theme). Your hack would always set the background for universal images when on the dark background even though they should be readable fine?

lukaw3d: yep yep

PR #361, opened 2023-02-02
Revert "docs: Add BIOS settings"

amela:
I think I didn't squash correctly


kostko:
Yeah but merging the revert will not remove any commits, so this is not needed :)

PR #440, opened 2023-05-15
Add LICENSE file (CC BY 4.0+Apache 2.0)

tjanez:
This PR:
- adds Apache 2.0 to LICENSE
- adds CC BY 4.0 to LICENSE-docs
- adds License notice in footer
- adds License notice in README
- reorganizes footer (#557)
- removes the Docusaurus example leftovers from the static/img folder.


matevz (commented):
I'm not sure, whether apache is ok for the documentation-like content and images.

E.g. wikipedia doesn't mention it as compatible: https://en.wikipedia.org/wiki/Wikipedia:Image_use_policy

This would need some investigation.

Perhaps we should be generous and hand out all our docs, examples and images as public domain?

kostko:
Yeah it should probably be one of the [CC ones](https://creativecommons.org/share-your-work/)?

matevz:
I revisited this again and I can't seem to find the right combination of licenses. In short:
- documentation, images, figures and other docs components should be licensed so that we are compatible with Wikipedia's CC BY-SA (uploading the core illustrations to wikimedia commons sounds like a good idea). I think BY-SA is too strict for us and would rather just go with public domain CC0 license or CC BY, allowing usage of our docs in closed-source environments too (I'm having a random person doing survey on c10l chains in mind and wants to include our Oasis Architecture illustration in his work).
- all our examples (also code snippets) should be licensed under the no-warranty + patent retention license. This means Apache v2 or similar.
- BUT: Apache v2 is not compatible with CC BY-SA, see the [answer here](https://opensource.stackexchange.com/a/12481).

Whatever we decide on should also be reflected in oasis-core, oasis-sdk, cli and other repositories which we combine the docs from.

I suggest we:
- license this docs repo under CC BY
- put a clause in oasis-core, oasis-sdk and CLI readme that the:
  - material under docs/ folder is licensed under CC BY
  - the code itself is still licensed under Apache v2, but individual code snippets may be used freely for educational purposes
- update the docs.oasis.io footer claiming the site content is "(C) 2019-2023 Oasis Protocol Foundation, CC BY 4.0"

matevz:
I find a nice example of the proposed licensing plan in the [docusaurus](https://github.com/facebook/docusaurus) project itself. I think taking https://github.com/oasisprotocol/docs/pull/440#issuecomment-1750359151 is the right path forward!

image
image

kostko:
Sounds like a good plan!

matevz:
Ok, this is ready for review now. I got the inspiration for the copyright footer from the [LibreOffice](https://www.libreoffice.org/) website:
image

README.md
```diff
@@ -242,3 +242,11 @@ and their usage including the articles:
...
   We strongly suggest that you set up your own Web3 gateway for your Sapphire
   endpoint.
+
+## License
+
+The Oasis documentation software and code snippets inside the documentation are
+licensed under [Apache 2.0](./LICENSE).
+
+The Oasis documentation (the `/docs` folder) including the media (e.g. images
+and diagrams) is licensed under [Attribution 4.0 International](./LICENSE-docs).
```
kostko:
```suggestion
and diagrams) is licensed under [CC Attribution 4.0 International](./LICENSE-docs).
```

docusaurus.config.js
```diff
@@ -146,28 +146,90 @@ const config = {
...
+              {
+                label: 'Discord',
+                href: 'https://oasis.io/discord',
+              },
               {
                 label: 'Youtube',
                 href: 'https://www.youtube.com/channel/UC35UFPcZ2F1wjPxhPrSsESQ',
               },
+              {
+                label: 'Rosegarden',
```
kostko: This is a community-run project, I'm not sure we should link it in the footer.

README.md
matevz: I'm uncertain. According to https://chooser-beta.creativecommons.org/, the copyright notice looks like:
image

matevz: I agree, let's add the Create Commons term as titled in https://creativecommons.org/licenses/by/4.0/legalcode.txt

docusaurus.config.js
matevz: Replaced with X.

matevz: I replaced it with another GitHub link.

kostko (approved):

PR #447, opened 2023-05-26
Update hardware-recommendations

amela:
10GB is not enough


docs/node/run-your-node/prerequisites/hardware-recommendations.md
```diff
@@ -83,8 +83,8 @@ that use a trusted execution environment (TEE).
   * Recommended: 8 GB of ECC RAM
 
 * All ParaTime compute or client nodes:
```
matevz:
```suggestion
* Emerald, Sapphire or Cipher compute or client node (figures below are per ParaTime):
```

matevz (approved):
I hope that the new values are just. Also update the introductory sentence please.

docs/node/run-your-node/prerequisites/hardware-recommendations.md
gw0: If we start listing ParaTimes explicitly, then we should probably list them all over the place?

gw0 (approved):

docs/node/run-your-node/prerequisites/hardware-recommendations.md
matevz: Can you be more specific? CPU and storage requirements are already specific for each ParaTime. The existing `All ParaTime compute or client nodes` in the Memory section is misleading, because you can always [build your own ParaTime](https://docs.oasis.io/paratime/) where memory requirements could be very small (e.g. the key manager ParaTime).

gw0: You are right.

PR #454, opened 2023-06-13
docs: Reword OPL introduction chapter

matevz:
Still WIP. I'd like to revamp the introductory chapter of OPL a bit and the Overview chapter starting off with the tutorial. Currently, something seems out of place as if OPL is something for DAO. Actually the power of OPL is hidden in the bridge and seamless integration with it and we don't mention this anywhere. We should briefly describe how the bridge works and more importantly what's the "default" flow for dApp developers and where they can tweak it e.g. picking some other bridge than Celer, where to pay for the transaction relaying gas fee etc.

### DIFF
--- a/docs/dapp/opl/README.md
+++ b/docs/dapp/opl/README.md
@@ -1,8 +1,8 @@
 # Oasis Privacy Layer
 
-The Oasis Privacy Layer (OPL) is an EVM-compatible privacy solution that opens
-new frontiers for confidential smart contracts in Web3 applications on the
-most popular EVM networks like Ethereum, BNB Chain, and Polygon.
+The Oasis Privacy Layer (OPL) is an EVM-compatible privacy solution that allows
+you to extend your existing contracts running on Ethereum, BNB Chain, and
+Polygon with secret transactions and state.
 
 ## DApp Development
 


docs/dapp/opl/README.md
```diff
@@ -1,8 +1,8 @@
 # Oasis Privacy Layer
 
-The Oasis Privacy Layer (OPL) is an EVM-compatible privacy solution that opens
-new frontiers for confidential smart contracts in Web3 applications on the
-most popular EVM networks like Ethereum, BNB Chain, and Polygon.
+The Oasis Privacy Layer (OPL) is an EVM-compatible privacy solution that allows
+you to extend your existing contracts running on Ethereum, BNB Chain, and
```
aefhm, line 4: I lifted the first draft from the marketing [copy](https://oasisprotocol.org/opl) 😬 

Maybe the sentiment should be more ambitious?

docs/dapp/opl/README.md
```diff
@@ -1,8 +1,8 @@
...
+Polygon with secret transactions and state.
```
aefhm, line 5:
```suggestion
Polygon with encrypted transactions and confidential state.
```

matevz:
Superseeded https://github.com/oasisprotocol/docs/pull/461.

PR #462, opened 2023-06-27
dApps > Sapphire > Quickstart - Restrict hardhat version to 2.12.x

mirayashi:
The package `@oasisprotocol/sapphire-hardhat` has been designed for hardhat 2.12.x. Following the docs exactly as written results in a non-functional `run-vigil.ts` in the end, since `npx hardhat` downloads the latest hardhat version, which is 2.16.0 at the time of writing this.

This pull request specifies the version of hardhat to initialize so following the instructions works as intended.

### DIFF
--- a/docs/dapp/sapphire/quickstart.mdx
+++ b/docs/dapp/sapphire/quickstart.mdx
@@ -289,7 +289,7 @@ Let's make it happen!
 We're going to use Hardhat this time because it's very convenient to use.
 
 1. Make & enter a new directory
-2. `npx hardhat` then create a TypeScript project.
+2. `npx hardhat@~2.12.0` then create a TypeScript project.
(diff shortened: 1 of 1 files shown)


aefhm (approved):
Thanks for the PR!

aefhm:
[`2.12.7`](https://www.npmjs.com/package/hardhat/v/2.12.7) is out but still behind. We need to fix this via https://github.com/oasisprotocol/sapphire-paratime/issues/139.

mirayashi:
Yeah ideally the sapphire hardhat package should be updated in order to support the most recent version, updating the docs with a specific version should be considered a short term solution. Btw I used the syntax `@~x.y.z` so it only pins `x.y` and takes whatever is the latest `z`, so here `2.12.7` indeed.

matevz:
LGTM, thanks!

PR #471, opened 2023-07-14
docs: Reorganize landing page to expose Sapphire and OPL

matevz:
- Moves "Create dApp" right after the "Use Oasis" in the top toolbar
- Revamps the landing page to include Oasis Sapphire at the top (right after the general info) followed by OPL.
- Replaces "Overview" chapter title with concrete "Use Oasis", "Create dApp", "Run Node" or "Build ParaTime" since this reflects the window/tab title and should be unique. The "Overview" toolbar label in the sidebar is still shown.
- Bumps ADRs repo to also include the new ADR 21 and 22.
- Website's general "Oasis Network Documentation" title is now just "Oasis Documentation"
- Some Run Node chapter cosmetic fixes

### DIFF
--- a/docs/README.mdx
+++ b/docs/README.mdx
@@ -1,3 +1,4 @@
+import DocCard from '@theme/DocCard';
 import DocCardList from '@theme/DocCardList';
 import {findSidebarItem} from '@site/src/sidebarUtils';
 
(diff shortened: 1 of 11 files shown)


kostko (approved):
Nice!

docs/README.mdx
```diff
@@ -1,18 +1,46 @@
+import DocCard from '@theme/DocCard';
 import DocCardList from '@theme/DocCardList';
 import {findSidebarItem} from '@site/src/sidebarUtils';
 
-# Getting Started
+# Oasis Platform Documentation
```
tjanez: This is actually just "Oasis Docs".

But perhaps not to duplicate, we could just have an inviting paragraph along the lines of:
Welcome to Oasis... This our documentation... Explore if further...

docs/README.mdx
```diff
@@ -1,18 +1,46 @@
...
+distinction between the consensus layer and different ParaTimes. We will also
+cover wallets and other tools for managing your assets across the Oasis chains
+and use unique Oasis features.
 
 <DocCardList items={[
     findSidebarItem('/general/oasis-network/'),
     findSidebarItem('/general/manage-tokens/'),
 ]} />
 
+## DApp Developers
```
tjanez: I would rename this to "Create dApp" to be consistent with the above header.
And we also avoid the ambiguous "DApp" vs. "dApp" capitalization.

docs/README.mdx
```diff
@@ -1,18 +1,46 @@
...
+platform is best known for confidentiality, the most notable ParaTime is
+[Oasis Sapphire], an EVM-compatible ParaTime with built-in contract state
+encryption. The Oasis team also prepared a set of libraries called the
+[Oasis Privacy Layer] to bridge existing dApps running on other chains to use
+the unique Sapphire's confidentiality features.
+
+The section also covers other ParaTimes such as the non-confidential
+[Oasis Emerald] and Wasm-compatible, confidential [Oasis Cipher].
+
+<DocCard item={ findSidebarItem('/dapp/') } />
```
tjanez: I would just omit this huge card. To have more focus on Sapphire and OPL.

docs/README.mdx
```diff
@@ -40,28 +68,24 @@ queries on the network.
     findSidebarItem('/node/run-your-node'),
 ]} />
 
-## DApps and ParaTime Developers
+## ParaTime Developers
```
tjanez: Can you change this to "Build ParaTime" (and similarly for other headlines).

docs/README.mdx
matevz: The main title is a requirement for markdown documents. Keep in mind that this also becomes the title of the browser window/tab, so it should be something obvious to make it easy when you're switching between the tabs.

matevz: Actually, let's revert this - the "Getting Started" seems fine since the "Oasis Documentation" is also appended to the title of each chapter.

PR #475, opened 2023-07-24
Delete bitpie-wallet.md

Kohlivish:
We should delete this as Bitpie doesn't support ROSE now (Announcement - https://medium.com/bitpie/announcement-on-suspension-of-support-for-algo-and-rose-fc35cb322617)


matevz:
Superseeded by #483 .

PR #476, opened 2023-07-25
sapphire: initial notes on authenticated calls & contract authentication

CedarMist:
This covers patterns for verifying caller identity for view calls.

Preview: https://deploy-preview-476--trusting-archimedes-14c863.netlify.app/dapp/sapphire/authentication

PR #481, opened 2023-07-31
Sapphire/Precompiles: add notes about <=0.5.2 eth_call RNG behavior + detail on how VRF works

CedarMist:
image


docs/dapp/sapphire/precompiles.md
```diff
@@ -60,14 +60,33 @@ Feel free to discover other convenient libraries for Solidity inside the
...
-domain separation when needed.
+Generate `num_bytes` pseudo-random bytes, with an optional personalization
+string (`pers`) added into the hashing algorithm to increase domain separation
+when needed.
 
 ```solidity
 bytes memory randomPad = Sapphire.randomBytes(64, "");
 ```
 
+:::danger Prior to 0.5.3
```
kostko: This will be 0.6.0.

docs/dapp/sapphire/precompiles.md
```diff
@@ -60,14 +60,33 @@ Feel free to discover other convenient libraries for Solidity inside the
...
+Sapphire.randomBytes(64, abi.encodePacked(msg.sender, this.perContactSecret));
+```
+:::
+
+The mode (e.g. simulation or 'view call' vs transaction execution) is fed to TupleHash (among other
+block-dependent components) to derive the "key id", which is then used to derive a per-block VRF key
+from epoch-ephemeral entropy (using KMAC256 and cSHAKE) so a different "key id" will result in a
+unique per-block VRF key. This per-block VRF key is then used to create the per-block root RNG which
+is then used to derive domain-separated (using Merlin transcripts) per-transaction random RNGs which
+are then exposed via this precompile.
```
kostko: Maybe move this under its own `### Implementation Details` or similar. Would also be nice if you could find links for KMAC256, cSHAKE, Merlin and TupleHash.

kostko (commented):
Only reviewing the part that is not based on #476, I suggest you remove the duplicate commits from this PR.

docs/dapp/sapphire/precompiles.md
```diff
@@ -60,14 +60,33 @@ Feel free to discover other convenient libraries for Solidity inside the
...
+
+This does not affect transactions, so if your contract requires confidentiality please
+generate a secret in the constructor and use it from view calls.
+
+```solidity
+Sapphire.randomBytes(64, abi.encodePacked(msg.sender, this.perContactSecret));
+```
+:::
+
+The mode (e.g. simulation or 'view call' vs transaction execution) is fed to TupleHash (among other
```
aefhm, line 84: (nice to have) I wish we had a diagram/design doc visual for this.

docs/dapp/sapphire/precompiles.md
```diff
@@ -60,14 +60,33 @@ Feel free to discover other convenient libraries for Solidity inside the
...
 
 ```solidity
 bytes memory randomPad = Sapphire.randomBytes(64, "");
 ```
 
+:::danger Prior to 0.5.3
+All view queries and simulated transactions (via `eth_call`) would receive the
+same entropy in-between blocks if they use the same `num_bytes` and `pers` parameters.
+
+This does not affect transactions, so if your contract requires confidentiality please
```
aefhm: > This does not affect transactions

As in, transactions are not able to use this precompile?

CedarMist: I will clarify, both transactions and view calls can run this.

But each transaction gets its own RNG seed.

While view calls all use a common per-block RNG seed. Meaning if two contracts request 32 bytes of random data with the same personalisation seed, they will get the same data. That could reveal sensitive info etc.

CedarMist:
> Only reviewing the part that is not based on #476, I suggest you remove the duplicate commits from this PR.

This has been rebased onto `main`. In future I will not use a fork, but a branch of this repo for changes.

kostko (approved):

PR #508, opened 2023-08-24
docs/dapp/sapphire: Add Contract verification section

matevz:
[Preview](https://deploy-preview-508--trusting-archimedes-14c863.netlify.app/dapp/sapphire/guide#contract-verification)


CedarMist:
Is it worth having this in a separate page, rather than in the Guide?

CedarMist (approved):

matevz:
> Is it worth having this in a separate page, rather than in the Guide?

Since it doesn't have any subtitles, let's leave it as a part of the guide for now.

PR #527, opened 2023-09-07
docs: Update SGX Device Permissions

amela:
Preview: https://deploy-preview-527--trusting-archimedes-14c863.netlify.app/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee#ensure-proper-sgx-device-permissions


docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
```diff
@@ -141,6 +141,15 @@ sudo adduser oasis sgx
 
 Failure to do so may result in permission denied errors during runtime startup.
 
+Alternatively, if you want to modify the permissions of `/dev/sgx_enclave`
```
kostko, line 144: When would this be preferred? I think adding the user to the right group seems like the better solution? Or is this for older distributions that didn't set up the `sgx` group?

amela: Yes, for older distributions that did not set up the sgx group.

PR #536, opened 2023-09-14
docs/node/testnet: Add notice about prebuild binaries

amela:


docs/node/testnet/upgrade-log.md
```diff
@@ -67,6 +67,12 @@ an updated ParaTime containing new trust roots is released and adopted.
 
 7. Replace the old version of Oasis Node with version [23.0].
 
+:::info
+
+Oasis Core 23.0 binary is built only for Ubuntu 22.04. You'll have to build the core if you're using prior Ubuntu versions.
```
kostko:
```suggestion
The Oasis Core 23.0 binary in our published releases is built only for Ubuntu 22.04. You'll have to build it yourself if you're using prior Ubuntu versions (or other distributions using older system libraries).
```

PR #546, opened 2023-09-23
feat: add flair indexing

kasrakhosravi:
Add Flair (https://flair.dev/) as an indexing solution to Oasis docs


matevz:
Out of curiosity how far would Flair support Sapphire? Emerald is "yet another EVM" while Sapphire unwinds the true potential of Oasis as a privacy-preserving platform.

kasrakhosravi:
> Out of curiosity how far would Flair support Sapphire? Emerald is "yet another EVM" while Sapphire unwinds the true potential of Oasis as a privacy-preserving platform.

Great question. so for supporting "Sapphire" indexing, all is needed is an RPC endpoint. I can create another docs entry for supporting indexing on Sapphire also?

PR #612, opened 2023-11-14
docs/node: Add upgrade epoch

amela:


docs/node/mainnet/eden-upgrade.md
```diff
@@ -148,7 +137,7 @@ upgrade epoch.
 3. Verify the provided Mainnet upgrade genesis file by comparing it to the
    local network state dump.
 
-   Find the `genesis-mainnet-TBD.json` file in the `exports` subdirectory in
+   Find the `genesis-mainnet-2023-11-29.json` file in the `exports` subdirectory in
```
kostko: That will not be the case, it will be `genesis-oasis-3-at-HEIGHT.json`, where HEIGHT corresponds to the block height at which the snapshot was taken.

kostko (approved):

PR #636, opened 2023-11-24
docs: Migrate Sapphire docs to sapphire-paratime repo

matevz:
This PR:
- adds oasisprotocol/sapphire-paratime git module.
- migrates Sapphire docs from this repo to sapphire-paratime including Sapphire-related images and diagrams.
- updates remark plugins for URL transformations.

Merge after https://github.com/oasisprotocol/sapphire-paratime/pull/227 is merged in.

[Preview](https://deploy-preview-636--trusting-archimedes-14c863.netlify.app/dapp/sapphire/)


CedarMist:
Out of curiosity, why are the individual .md files symlinked, instead of the sapphire/docs directory being symlinked?

matevz:
> Out of curiosity, why are the individual .md files symlinked, instead of the sapphire/docs directory being symlinked?

I thought keeping README.mdx inside the docs repo might be a good idea, because it contains stuff like endpoint URLs and Metamask network installation snippets.

So this prevents you from linking the whole directory, because you cannot inject individual files in then.

btw. We use the same approach in Cipher. To update docs.oasis.io with the new version of the docs inside the external repo you need to bump the commit to external repo (we use dependabot for this—you just need to approve the PR). If you add a new file, you also need to add a new symlink inside the docs folder then (and add redirection, if you removed anything).

CedarMist (approved):
Ok this seems fine, dependabot will submit PR on new Sapphire releases (but not releases of contracts or clients) and we can just remember to periodically update the submodule and add/remove individual .md files as necessary (less ideal, but it works for now)

PR #645, opened 2023-11-28
docs/node/sentry: Update config file for 23.0

amela:
Migrated with `oasis-node config migrate` and updated variables. Please review.

### DIFF
--- a/docs/node/run-your-node/sentry-node.md
+++ b/docs/node/run-your-node/sentry-node.md
@@ -56,86 +56,49 @@ Before using this configuration you should collect the following information to 
   on the validator node. Note that the above command is only available in `oasis-node` from version 20.8.1 onward.
 
 ```yaml
-##
-# Oasis Sentry Node Configuration
-#
-# This file's configurations are derived from the command line args found in the
-# root command of the oasis-node binary. For more information execute:
-#
-#     oasis-node --help
-#
-# Settings on the command line that are separated by a dot all belong to the
-# same nested object. So "--foo.bar.baz hello" would translate to:
-#
-#     foo:
-#       bar:
-#         baz: hello
-##
-
-# Set this to where you wish to store node data. The node artifacts
-# should also be located in this directory (for us this is /node/data)
-datadir: /node/data
-
-# Logging.
-#
-# Per-module log levels are defined below. If you prefer just one unified log
-# level, you can use:
-#
-# log:
-#   level: debug
-log:
-  level:
-    # Per-module log levels. Longest prefix match will be taken. Fallback to
-    # "default", if no match.
-    default: debug
-    tendermint: warn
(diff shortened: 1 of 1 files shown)


docs/node/run-your-node/sentry-node.md
```diff
@@ -225,78 +146,33 @@ Before using this configuration you should collect the following information to
...
-  registration:
-    # In order for the node to register itself the entity.json of the entity
-    # used to provision the node must be available on the node.
-    entity: /node/etc/entity.json
-  sentry:
-    address:
-      - "{{ sentry_node_grpc_public_key }}@{{ sentry_node_private_ip }}:9009"
-
-# Consensus backend.
+mode: client
```
ptrus: There were previously two configs, one for the sentry node, and one for the "normal" node, which is connecting to the sentry. Seems that one of the configs is missing now?

amela: Yeah, you are right. Added both configs and re-added comments.

ptrus (approved):
I think this should work (didn't test it atm though)

docs/node/run-your-node/sentry-node.md
```diff
@@ -225,78 +188,44 @@ Before using this configuration you should collect the following information to
...
+        # the default
+        #file: /var/log/oasis-node.log
+
+consensus:
+    listen_address: tcp://0.0.0.0:26656
+    p2p:
+        disable_peer_exchange: true
+        persistent_peers:
+            - {{ sentry_node_tendermint_id }}@{{ sentry_node_private_ip }}:26656
+    prune:
```
ptrus: Does anyone know why did we have pruning configured here for the validator? e.g. for normal validator node we don't have pruning configured in the examples. I would just remove this "prune section" from this config (i think it's ok to keep in the other (sentry) config above, it makes more sense there).

---
edit actually idk. if it makes sense on the sentry to keep it this low. That means if validator is disconnected for ~1 hour it won't be able to sync from the sentry? Maybe remove prunning there as well, since it's not necessary.

docs/node/run-your-node/sentry-node.md
```diff
@@ -56,86 +56,49 @@ Before using this configuration you should collect the following information to
...
+            # Per-module log levels. Longest prefix match will be taken. Fallback to
+            # "default", if no match.
+            default: debug
+            # By default logs are output to stdout. If you're running this in docker 
+            # keep the default
+            #file: /var/log/oasis-node.log
+consensus:
+    external_address: tcp://{{ external_address }}:26656
+    listen_address: tcp://0.0.0.0:26656
+    prune:
```
ptrus: I would remove pruning here as well since it's not really necessary and we usually don't have it configured in these example configs.

PR #677, opened 2023-12-07
Mention finality in Sapphire README.mdx

CedarMist:


kostko (approved):

matevz:
This is probably true for Emerald and Cipher too right?

kostko:
Yeah.

PR #743, opened 2024-02-21
Update paratime-node.mdx p2p.port config path and default value

DanielMagro97:
Updated p2p.port config path and default value

### DIFF
--- a/docs/node/run-your-node/paratime-node.mdx
+++ b/docs/node/run-your-node/paratime-node.mdx
@@ -294,12 +294,12 @@ ias:
 
 p2p:
     # External P2P configuration.
-    port: 30002
+    port: 9200
     registration:
         addresses:
             # The external IP that is used when registering this node to the
             # network.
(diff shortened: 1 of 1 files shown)


docs/node/run-your-node/paratime-node.mdx
```diff
@@ -332,7 +332,7 @@ Before using this configuration you should collect the following information to
 
 :::caution
 
-Make sure the `worker.p2p.port` (default: `9200`) port is exposed and publicly
+Make sure the `p2p.port` (default: `30002`) port is exposed and publicly
```
kostko: Ah, actually I just checked and 9200 is still the default port (if not set in config), it is just that the above example sets it to 30002. So maybe to make it more clear this should be "defaults to `9200` if not set in config"?

DanielMagro97: Updated the port in the sample config and note to 9200

kostko (approved):

PR #755, opened 2024-03-06
docs: Use recommended replacement for deprecated apt-key command

uscinski:
Section on Trusted Execution Environments uses the `apt-key` command, which is deprecated and will be removed in future Ubuntu and Debian versions. Here we use recommended commands and directory locations to add a trusted repository.

### DIFF
--- a/docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
+++ b/docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
@@ -108,8 +108,8 @@ is to use the Intel's [official Intel SGX APT repository](https://download.01.or
 First add Intel SGX APT repository to your system:
 
 ```bash
-echo "deb https://download.01.org/intel-sgx/sgx_repo/ubuntu $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/intel-sgx.list >/dev/null
-curl -sSL "https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key" | sudo -E apt-key add -
+curl -fsSL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor -o /usr/share/keyrings/intel-sgx-deb.gpg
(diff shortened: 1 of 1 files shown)


kostko (commented):
Ew, this is more complicated than before so more likely something will go wrong :(

kostko:
But I guess this can't be avoided?

uscinski:
> Ew, this is more complicated than before so more likely something will go wrong :(

This is indeed more complex. It follows the [standardized procedure on how to configure an unofficial (i.e., a third-party) APT repository on a machine](https://wiki.debian.org/DebianRepository/UseThirdParty). From the proposed standard it differs in that it _does not_ perform standard pinning, which would make the entire procedure even more verbose. For example, a similar procedure can be spotted in the [official Docker documentation](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository).

> But I guess this can't be avoided?

It _can_ be avoided. A drop-in replacement for the `apt-key add` command is still a one-liner. Let's show this on an  example from the man pages.

With `apt-key`:

```bash
wget -qO- https://myrepo.example/myrepo.asc | sudo apt-key add -
```
Without `apt-key`, using ASCII armored key:

```bash
wget -qO- https://myrepo.example/myrepo.asc | sudo tee /etc/apt/trusted.gpg.d/myrepo.asc
```

Without `apt-key`, using key in the binary OpenGPG format:

```bash
wget -qO- https://myrepo.example/myrepo.gpg | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/myrepo.gpg
```

To keep it simple, we can go with the last example.

However, placing keys in `/etc/apt/trusted.gpg.d` still results in a warning on recent Debian systems, because APT completely and unconditionally trusts all the keys stored there, and would therefore accept _all_ the packages signed by _any_ of the stored keys, regardless of a repository the packages (or the key) originate from. To limit the scope of keys from the unofficial repositories, it is therefore recommended to store them into `/etc/apt/keyrings`, which is a directory that _might not_ exist on a system, and refer to them with the `signed-by` directive in the respective `sources.list` entries. There, for complex entries it is recommended to use the "Deb822" file format.

Edit: @kostko Using the same assumptions as before, I have now simplified the code.

kostko (approved):

PR #774, opened 2024-03-27
docs: Update staking rewards graph

amela:
Preview: https://deploy-preview-774--oasisprotocol-docs.netlify.app/general/oasis-network/token-metrics-and-distribution#staking-incentives


matevz (changes requested):
Please add a record to the Changelog section at the end of this page. Something like:
```
- Mar 28, 2024:
  - Updated Staking Rewards Schedule based on the [Staking parameters change proposal](https://www.oasisscan.com/proposals/4) passed Mar 26 2024.
```

kostko (commented):
Also update the change log section to specify that the reward schedule has been adjusted after governance proposal 4 passed which changed the schedule.

docs/general/oasis-network/token-metrics-and-distribution.mdx
```diff
@@ -103,3 +103,7 @@ For more details, see its [Delegation Policy](../../get-involved/delegation-poli
 * **Nov 2, 2020:**
   * Updated Backers image to include more publicly-known backers.
   * Included a community-proposed (and foundation supported) increase in staking rewards range from 15% - 2% to 20% - 2% over the first four years of the network. Impacted charts (distribution, token delivery schedule, and expected staking rewards) also updated to reflect the increase in staking rewards.
+* **Mar 27, 2024:**
```
kostko: This needs to be at the top.

docs/general/oasis-network/staking_rewards/data.csv
```diff
@@ -0,0 +1,353 @@
+Until epoch,Estimated date,Reward per epoch %,Estimated annualized rewards %
```
lukaw3d: duplicated file

src/staking_rewards/data.csv
```diff
@@ -350,97 +350,4 @@ Until epoch,Estimated date,Reward per epoch %,Estimated annualized rewards %
...
-36714,2024-11-16T22:50:26.626Z,0.000010,0.0898
-36738,2024-11-17T22:15:02.336Z,0.000008,0.0719
-36762,2024-11-18T21:39:38.047Z,0.000007,0.0629
-36786,2024-11-19T21:04:13.757Z,0.000006,0.0539
-36810,2024-11-20T20:28:49.467Z,0.000005,0.0449
-36834,2024-11-21T19:53:25.178Z,0.000004,0.0359
-36858,2024-11-22T19:18:00.888Z,0.000003,0.0269
-36882,2024-11-23T18:42:36.598Z,0.000002,0.0180
-36930,2024-11-25T17:31:48.019Z,0.000001,0.0090
+81662,2030-01-01T00:00Z,0.000283,2.5741
```
lukaw3d: data.json says until epoch 90000, not 81662

tjanez (commented):
The new graph looks too wide because it is effectively computed until year 2031.
Can we start by narrowing it to 2020 -> 2028?

amela:
@tjanez @lukaw3d, if we narrow it down to 2020 -> 2028, the data in the CSV will not match the JSON (90000). Is that okay? Until now, the JSON had the same data as the staking rewards schedule.

src/staking_rewards/data.csv
amela: I thought it would look nicer if the graph ended in 2030 rather than on a random date. Do the JSON and csv need to match?

lukaw3d: That doesn't make sense. CSV doesn't affect the graph

lukaw3d: Either way, don't manipulate the data, especially the data that users download and analyze and make own graphs out of. IF we want to display less data then we should change that in the display component :shrug:

and that can be a separate pullrequest if tadej insists on not letting users to zoom where they want to

amela: Yeah, this makes sense. I changed the CSV to include epoch 90000. We can create a new pr for the display component if needed. I also think it is nicer to show all the data and then can users to zoom in if they want.

kostko (approved):
I think this is good as it reflects the updated schedule. We can further adjust the chart in subsequent PRs.

matevz (approved):

PR #793, opened 2024-04-18
mention publicly exposing ports in node configuration pages

ptrus:
Mention both TCP and UDP (currently TCP is used for both, but in future QUIC could be used, so make this more future proof).

Probably should add the note also to the paratime-client-node and non-validator node? Since the network will benefit from more nodes having the p2p ports open, even if these are client nodes.


kostko (approved):
> Probably should add the node also to the paratime-client-node and non-validator node? Since the network will benefit from more nodes having the p2p ports open, even if these are client nodes.

Agreed.

PR #818, opened 2024-05-13
docs: Add Sapphire deployment pattern section

aefhm:
## Description

Add necessary file reference. Blocked by https://github.com/oasisprotocol/sapphire-paratime/pull/313.

### DIFF
--- a/docs/dapp/sapphire/deployment.md
+++ b/docs/dapp/sapphire/deployment.md
@@ -1,0 +1,1 @@
+../../../external/sapphire-paratime/docs/deployment.md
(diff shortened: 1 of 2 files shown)


matevz:
You can also bump the external/sapphire-paratime repo in this PR so that the new proxy.md will actually be there and tests will pass.

aefhm:
> You can also bump the external/sapphire-paratime repo in this PR so that the new proxy.md will actually be there and tests will pass.

That is after merging the other PR right? Submodule config looks set to `main`.

Edit. I think I figured it out.

matevz (approved):
Yep, looks correct ;)

PR #854, opened 2024-06-17
docs/node: Document 'Enclave panicked' troubleshooting for ParaTime Node

anzoman:
Here, we describe the SGX troubleshooting for the _Enclave panicked_ error when running ParaTime Node.

Connected to https://github.com/oasisprotocol/tools/pull/36 which adds a check for AES-NI support to `attestation-tool`.

### DIFF
--- a/docs/node/run-your-node/paratime-node.mdx
+++ b/docs/node/run-your-node/paratime-node.mdx
@@ -413,6 +413,39 @@ enough.
 
 Double check your node entity satisfies the staking requirements for a ParaTime node. For details see the [Stake Requirements](paratime-node.mdx#stake-requirements) section.
 
+### Enclave panicked
+
+If there is a misconfiguration in the prerequisite [BIOS settings], you can see an error in the logs reporting a
+problem when running SGX enclaves.
+
+```json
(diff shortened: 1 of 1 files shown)


docs/node/run-your-node/paratime-node.mdx
```diff
@@ -413,6 +413,39 @@ enough.
...
+```
+AES instruction                         = true
+```
+
+If the AES instruction is set to `false`, you need to reconfigure you BIOS and set it to `true`.
+You can do similar inspection for other [BIOS settings].
+
+:::tip
+
+You can use the [attestation tool] that also checks if the AES instruction set is available.
```
ptrus: Let's add a note that or somehow update the text so that it mentions that at least version 0.3.4 of attestation-tool is needed.

anzoman: Added.

docs/node/run-your-node/paratime-node.mdx
```diff
@@ -413,6 +413,39 @@ enough.
...
+```
+AES instruction                         = true
+```
+
+If the AES instruction is set to `false`, you need to reconfigure you BIOS and set it to `true`.
+You can do similar inspection for other [BIOS settings].
+
+:::tip
+
+You can use the [attestation tool](at least version 0.3.4) that also checks if the AES instruction set is available.
```
ptrus:
```suggestion
You can use the [attestation tool] (at least version 0.3.4) that also checks if the AES instruction set is available.
```

ptrus (approved):
Looks good 👍

docs/node/run-your-node/paratime-node.mdx
anzoman: Added space.

PR #869, opened 2024-07-04
Handle repeated AddToMetaMask clicks

lukaw3d:
If user already has a chain added in metamask, then button does nothing. Added a message

Before, After:

[before,after.webm](https://github.com/oasisprotocol/docs/assets/3758846/a8cacfcf-a2c1-47ef-b2f2-de2a7ec8f90c)

### DIFF
--- a/src/AddToMetaMask.tsx
+++ b/src/AddToMetaMask.tsx
@@ -18,6 +18,7 @@ const AddToMetaMask = (props: {
           if (!window.ethereum?.request) {
             return alert('Have you installed MetaMask yet? If not, please do so.\n\nComputer: Once it is installed, you will be able to add the ParaTime to your MetaMask.\n\nPhone: Open the website through your MetaMask Browser to add the ParaTime.')
           }
+          const startTime = Date.now()
           window.ethereum.request({
             method: 'wallet_addEthereumChain',
             params: [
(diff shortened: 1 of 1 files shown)


src/AddToMetaMask.tsx
```diff
@@ -33,6 +33,8 @@ const AddToMetaMask = (props: {
                 blockExplorerUrls: props.be,
               },
             ],
+          }).then(response => {
+            if (response === null) alert('Added')
```
matevz: Can you change the alert to something like:
```
The RPC endpoint for Chain ID {props.chainId} already exists.
If you want to replace it, please remove the old RPC endpoint first and then add the new one.
```

```suggestion
            if (response === null) alert('Added')
```

matevz (approved):
Good idea. I suggest a slightly more sophisticated message.

src/AddToMetaMask.tsx
lukaw3d: - This message is shown when you first add Sapphire too. We would need to measure how quickly null is returned to differentiate "already exists" vs "added first time".
- If you have OPF sapphire RPC added and you click 1RPC: metamask will ask you if you want to add 1RPC as a separate entry. No need to remove first. (if we really wanted to, I guess we could detect any existing entries with wallet_switchEthereumChain)

src/AddToMetaMask.tsx
```diff
@@ -33,6 +34,8 @@ const AddToMetaMask = (props: {
                 blockExplorerUrls: props.be,
               },
             ],
+          }).then(response => {
+            if (response === null && Date.now() - startTime < 100) alert('Already added')
```
matevz: How about:
```
The {props.name} RPC already added.
```

matevz (approved):

src/AddToMetaMask.tsx
matevz: Maybe also add
```
// XXX: A hack to figure out, if the RPC already exists.
```

lukaw3d: +1

PR #876, opened 2024-07-08
docs/node: Reorder paratimes and minor fixes

kostko:


kostko:
Also changed the example hostname(s) used in the gRPC node guide to use `example.com` as per the RFC.

docs/node/mainnet/README.md
```diff
@@ -56,6 +56,27 @@ for the current Testnet parameters.
 
 This section contains parameters for various ParaTimes known to be deployed on the Mainnet.
 
+### Sapphire
```
matevz, line 59: While you're at it, can you also swap the Emerald-Cipher menu entries in https://github.com/oasisprotocol/docs/blob/main/sidebarDapp.ts#L54-L83 ?

matevz (approved):

PR #904, opened 2024-08-22
set-up-tee: Mention logout/login for permissions to take effect

matevz:
Noticed I needed to logout/login to my computer for SGX permissions to take effect.

### DIFF
--- a/docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
+++ b/docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
@@ -53,7 +53,11 @@ and the user running Oasis Node is `oasis`, you can do:
 sudo adduser oasis sgx
 ```
 
-Failure to do so may result in permission denied errors during runtime startup.
+Failure to do so may result in the "permission denied OS error 13" during
+runtime startup.
+
+If you are planning to run your node from an interactive session, make sure to
+log out for permissions to take effect.
 
 ## AESM Service
 


docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
```diff
@@ -70,7 +70,11 @@ and the user running Oasis Node is `oasis`, you can do:
 sudo adduser oasis sgx
 ```
 
-Failure to do so may result in permission denied errors during runtime startup.
+Then, restart your computer or reload the SGX modules for permissions to take
```
kostko: Hm, you shouldn't need to restart your computer. However if you are running your node from an interactive session (e.g. a terminal with your user account), then you need to terminate the session (logout) and login again.

But if you are running your node via something like systemd, nothing like that should be needed.

matevz: > Hm, you shouldn't need to restart your computer. However if you are running your node from an interactive session (e.g. a terminal with your user account), then you need to terminate the session (logout) and login again.

Oops, you were right. Closing/Opening the terminal app didn't work, because they're probably forked from the kde session anyway, but logging out of the desktop and back in works. Will fix.

docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
```diff
@@ -70,7 +70,11 @@ and the user running Oasis Node is `oasis`, you can do:
 sudo adduser oasis sgx
 ```
 
-Failure to do so may result in permission denied errors during runtime startup.
+Then, log out from your session (desktop, ssh) and back in for permissions to
```
kostko: Maybe say something like "If you are planning to run your node from an interactive session, make sure to log out from your ...". Because if you are running via systemd or similar, no such thing is needed as the process manager initializes the session.

kostko (approved):

PR #1013, opened 2024-10-30
docs/dapp/opl: remove secret-ballot-example

### LINKED ISSUE #1011: Remove OPL Secret Ballot example from docs
This example is no longer necessary.
It's very long winded & the ping examples are much better.

Before:
image

After:
image

CedarMist:
Closes #1011

I think it's sufficient to have @rube-de's examples docs for CelerIM, OPL & Router protocol, rather than a much longer convoluted example which becomes out of date easily.


rube-de (approved):
looks good for me.

Maybe we can keep the example code in the demo-opl repository.

aefhm (approved):
Agree.

aefhm:
> looks good for me.
> 
> Maybe we can keep the example code in the demo-opl repository.

I'm also okay if we archive that one (be more transparent).

PR #1020, opened 2024-11-13
Bump Docusaurus to 3.6.1

### LINKED ISSUE #1018: Bump Docusaurus to 3.6.x

matevz:
Fixes https://github.com/oasisprotocol/docs/issues/1018

- bumps docusaurus to 3.6.1 and other dependencies used in the docusaurus template project
- bumps docusaurus-search-local
- bumps mermaid-cli to 0.45.0 which is based on mermaid 11.4.0
- this requires installing a separate puppetteer dependency and a new config file for it to handle some sandbox annoyances
- bumps node to >=20 required by the new mermaid-cli version


.github/workflows/ci-links.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '18.x'
+          node-version: '20.x'
```
pro-wh, line 22: could go to the newest LTS 22.x. in other places we use the oldest supported LTS to test for compatibility, but this is just us building our docs, so newest is fine I think

pro-wh (approved):
I didn't look at any release notes

.github/workflows/ci-links.yml
aefhm: Yeah, I don't see why we shouldn't update across the board as Node 18 is getting dropped soon.

package.json
```diff
@@ -43,23 +44,23 @@
       "not op_mini all"
     ],
     "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
+      "last 3 chrome version",
```
aefhm, line 47: 👍 Not even a year of Chrome development.

aefhm (approved):

.github/workflows/ci-links.yml
matevz: Netlify doesn't support node 22.x yet, only 20.x.

pro-wh: oh does netlify run the build for us?

kostko: Yeah.

matevz: > oh does netlify run the build for us?

The preview pages are rendered on netlify for us. They use a separate stack running Ubuntu 20.04 and Node version up to 20.x.

PR #1047, opened 2024-12-10
tee: Add instructions for configuring DCAP on IBM Cloud

anzoman:
After testing the SGX support on IBM Cloud we are adding the instructions for configuring DCAP on IBM Cloud.

### DIFF
--- a/docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
+++ b/docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
@@ -161,12 +161,16 @@ Some cloud providers require you to use their PCCS.
 - Alibaba Cloud: See the [Alibaba Cloud documentation] for details on configuring the quote provider. The
   documentation shows the required `sgx_default_qcnl.conf` changes.
 
+- IBM Cloud: See the [IBM Cloud documentation] for details on configuring the quote provider. The
+  documentation shows the required `sgx_default_qcnl.conf` changes.
+
 - Other cloud providers: If you are using a different cloud service provider, consult their
 specific documentation for the appropriate PCCS configuration and guidance on configuring the quote provider, or
 use one of the other PCCS options.
 
 [Azure documentation]: https://learn.microsoft.com/en-us/azure/security/fundamentals/trusted-hardware-identity-management#how-do-i-use-intel-qpl-with-trusted-hardware-identity-management
 [Alibaba Cloud documentation]: https://www.alibabacloud.com/help/en/ecs/user-guide/build-an-sgx-encrypted-computing-environment
+[IBM Cloud documentation]: https://cloud.ibm.com/docs/vpc?topic=vpc-about-attestation-sgx-dcap-vpc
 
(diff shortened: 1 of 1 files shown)


docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
```diff
@@ -161,12 +161,16 @@ Some cloud providers require you to use their PCCS.
 - Alibaba Cloud: See the [Alibaba Cloud documentation] for details on configuring the quote provider. The
   documentation shows the required `sgx_default_qcnl.conf` changes.
 
+- IBM Cloud: See the [IBM Cloud documentation] for details on configuring the quote provider. The
```
ptrus, line 164: This is just for the VM instances, right? 

I assume the bare metal instances don't need these changes to the config.

gw0:
> After testing the SGX support on IBM Cloud

Didn't we confirm that currently SGX DCAP attestation on IBM VMs does not work? It supposedly works only on IBM bare-metal instances, but did we confirm this? If yes, I would expect that this PR adds a note to the list of cloud providers that only bare-metal instances work.

docs/node/run-your-node/prerequisites/set-up-trusted-execution-environment-tee.md
anzoman: Yes, just the VMs, for the bare-metal it works out of the box.

anzoman:
> Didn't we confirm that currently SGX DCAP attestation on IBM VMs does not work?

Not entirely, the IBM Cloud team confirmed it works on their side, and stated the two tools (`sgx-detect` and `attestation tool`) are unexpectedly throwing the I/O error. We will probably need to run an actual node to confirm whether it works for real.

Also, this part of the docs never explicitly states that SGX works with IBM Cloud VMs, but just provides links to configure PCCS according to the docs. If we ever want to state that it works only for bare-metal, we should probably do it inside some support metrics (e.g., in https://docs.oasis.io/node/run-your-node/prerequisites/cloud-providers#known-providers where we already have a table with known providers).

gw0:
> We will probably need to run an actual node to confirm whether it works for real.

As far as I know we developed the `attestation-tool` exactly for this purpose. If the `attestation-tool` does not work, something is misconfigured and the node should also not work.

> this part of the docs never explicitly states that SGX works with IBM Cloud VMs

True, but adding these instructions for IBM Cloud and not mentioning where exactly is this supposed to work, you are implicitly suggesting it works on all IBM Cloud instances. Lets wait until we actually confirmed it works.

anzoman:
After performing an internal test of running an Oasis client node for Testnet with Sapphire ParaTime, we confirmed that running production enclaves works on the IBM Cloud Virtual Servers for VPC.

The https://github.com/fortanix/rust-sgx/issues/565 issue causing `sgx-detect` and `attestation-tool` to fail can then also be seen as valid. I've added an additional commit to our TEE troubleshooting docs.

kostko (approved):

PR #1053, opened 2024-12-12
tooling: Specify yarn version via packageManager

aefhm:
## Description

Corepack likes an explicit `packageManager`. I believe this is `1.22.22` for us for now.

Note. We could consider going up to `v4` as we are already on Node 20.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,5 +62,6 @@
   },
   "engines": {
     "node": ">=20.0"
-  }
+  },
+  "packageManager": "yarn@1.22.22"
 }


matevz:
Is this really needed?
[Node API docs](https://nodejs.org/api/packages.html#packagemanager) tells me the `packageManager` directive is still experimental.

aefhm:
> Node API docs

Not a must have. I think people using [`corepack`](https://nodejs.org/api/corepack.html) will get that insertion in their diffs.

matevz (approved):
Should we leave out the sha512 hash? It clutters the `package.json` file and I don't see security issues omitting it.

PR #1102, opened 2025-01-21
docs/node/mainnet: Bump Oasis Core to 24.3.1

amela:


kostko (approved):

kostko:
Actually it needs to be 24.3.1!

PR #1124, opened 2025-02-07
docs/node: Update client node configuration

peternose:


kostko (commented):
Mainnet still has 24.3.2, so we shouldn't update docs until that is at 25.0? Otherwise this is good.

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -200,14 +200,22 @@ runtime:
     # Paths to ParaTime bundles for all of the supported ParaTimes.
     paths:
         - {{ runtime_orc_path }}
+    # Configure ParaTimes to run outside of a Trusted Execution Environment.
```
gw0, line 203: Is this even needed? If no TEE is specified, it is automatically selected based on availability, with TDX and SGX taking precedence over ELF. If we want to just illustrate how a RONL can be optionally configured, we could show an commented out example? Would it make sense to also show an example of the `runtime.runtimes[].config` field?
```
# config:
#   estimate_gas_by_simulating_contracts: true
#   allowed_queries:
#     - all_expensive: true
```

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -200,14 +200,22 @@ runtime:
     # Paths to ParaTime bundles for all of the supported ParaTimes.
```
gw0, line 200: Are ROFL runtimes also called ParaTimes? It might be cleaner if we revision this "ParaTime" naming convention and start using the new more meaningful names, "runtime on-chain logic (RONL) (aka ParaTime/chain)" and "runtime off-chain logic (ROFL)".

gw0 (approved):

docs/node/run-your-node/paratime-client-node.mdx
peternose: Ups, missed these comments.

> Is this even needed? If no TEE is specified, it is automatically selected based on availability, with TDX and SGX taking precedence over ELF.

No, we don't auto check if the host supports TEE and then run runtime in it. We could implement this in the future. But currently, if runtime want's to run in SGX and your node doesn't support that, it will fail while provisioning the runtime.

docs/node/run-your-node/paratime-client-node.mdx
peternose: In the code we are using name runtime for a group of components, one of them being RONL. In the future, the Sapphire runtime should look like this:
- one RONL component (paratime or runtime in old days),
- multiple system ROFL components (included in the bundle, auto upgraded together with RONL),
- custom ROFL components (that talk to RONL and other ROFLs).

gw0 (approved):

PR #1172, opened 2025-03-11
docs: Add dark mode CSS to Mermaid SVGs

### LINKED ISSUE #1070: docs: Improve Mermaid diagram SVGs in dark mode
## Description

Relates to https://github.com/oasisprotocol/docs/issues/308.

Mermaid diagrams currently are exported to an SVG that is read inline which keeps a white background even in dark mode.

We would like to reduce the contrast and improve the UX for devs using dark mode. Mermaid does not currently support dark mode detecting SVGs through CSS, but does pass themes into exported SVG and also supports overrides.

[Themed images](https://docusaurus.io/docs/markdown-features/assets#themed-images) is another approach. I encountered some difficulties during the last attempt as inlined SVGs [lost some attributes](https://github.com/oasisprotocol/docs/issues/308#issuecomment-2544126892), but we can give it another try. Other approaches are welcome too.

aefhm:
## Description

Close #1070 and blocked by https://github.com/oasisprotocol/sapphire-paratime/pull/538

Given that switching to Mermaid extension is also finicky with detecting dark mode, I'm continuing down the current path of overriding `themeCSS` through the usage of cli tool. It appears we do need to order the new media query at the end per specificity.

Reviewer should also review color choices.

[Least pretty](https://deploy-preview-1172--oasisprotocol-docs.netlify.app/build/opl/hyperlane/), [this](https://deploy-preview-1172--oasisprotocol-docs.netlify.app/build/opl/hyperlane/pingpong-example), [this](https://deploy-preview-1172--oasisprotocol-docs.netlify.app/build/opl/opl-sdk/) and [better](https://deploy-preview-1172--oasisprotocol-docs.netlify.app/build/opl/router-protocol/pingpong-example)

## TODO

- [x] re-generate SVGs in CI
- [x] merge https://github.com/oasisprotocol/sapphire-paratime/pull/538
- [x] bump `sapphire-paratime`


docs/build/diagrams/opl-hyperlane-flow.mmd.svg
aefhm, line 1: This diagram has hard coded colors. Do we want to keep them in dark mode? cc @rube-de

matevz (approved):
My 5 cents:
- could you use no background color (or the one that matches the docs background)
- personally I would use some more colors in the dark theme. Currently looks like 5 shades of gray ;)

Also, for some reason the fonts are slightly bigger, but cut-off on my machine:
image.
Perhaps you need to let the CI to generate the images using ubuntu default web-safe fonts and copy the generated .svg artifacts over?

Otherwise looks good, thanks.

mermaid-theme.css
```diff
@@ -0,0 +1,91 @@
...
+    .messageLine0, line {
+        stroke: white !important; 
+        fill: white !important;
+    }
+    
+    .activation0 {
+        fill: #666666 !important;
+        stroke: white; 
+    }
+}
```
aefhm:
```suggestion
}

```

aefhm:
> * could you use no background color (or the one that matches the docs background)

Done.

> * personally I would use some more colors in the dark theme. Currently looks like 5 shades of gray ;)

Okay, just a little bit color.


> Also, for some reason the fonts are slightly bigger, but cut-off on my machine: image. Perhaps you need to let the CI to generate the images using ubuntu default web-safe fonts and copy the generated .svg artifacts over?

That was it.

rube-de (approved):

matevz (commented):
The black arrows on this diagram don't seem to look right when using dark theme https://deploy-preview-1172--oasisprotocol-docs.netlify.app/build/opl/hyperlane/
image

I think the right approach would be to use white lines similar to the sequence diagram like here https://deploy-preview-1172--oasisprotocol-docs.netlify.app/build/opl/hyperlane/pingpong-example

And the colors of the subgraph backgrounds should be made darker.

Edit: I just realized the flowchart diagram above has hardcoded CSS. This should be moved outside similar to what we have in the sequence diagrams.

rube-de:
> Edit: I just realized the flowchart diagram above has hardcoded CSS. This should be moved outside similar to what we have in the sequence diagrams.

Probably it was an error from me to use their mermaid code, the better way is to just use an image, like with the other external architectures.

aefhm:
> > Edit: I just realized the flowchart diagram above has hardcoded CSS. This should be moved outside similar to what we have in the sequence diagrams.
> 
> Probably it was an error from me to use their mermaid code, the better way is to just use an image, like with the other external architectures.

Ah, that sourcing makes sense. I could see how you may not want to mislead readers and claim ownership of this IP. That said, their site does support dark mode, so I can probably go either way. 🤔

aefhm:
Light
image

Dark
image

It isn't terrible if we just use their SVG.

docs/build/diagrams/opl-hyperlane-flow.mmd.svg
aefhm, line 1: I'll need to fix the width.

matevz:
> Light image
> 
> Dark image
> 
> It isn't terrible if we just use their SVG.

What if we simply removed CSS from the existing .mmd? Either is fine with me, let's get this PR merged.

rube-de:
> What if we simply removed CSS from the existing .mmd? Either is fine with me, let's get this PR merged.

Removing the css from the mmd will remove highlighting the hyperlane parts, which reduces the helpfulness of the diagram.

aefhm:
Merging to unblock the other `sapphire-paratime` bumps. Happy to do follow up.

Issue #1130, opened 2025-02-10
docs: Add basic analytics for docs.oasis.io


matevz:
Something like rose app that doesn't need user's consent.

lubej:
Note to self: We should talk about transfer of ownership of fathom analytics.

PR #1287, opened 2025-06-19
Add analytics

lubej:


docusaurus.config.ts
```diff
@@ -244,6 +244,10 @@ Apache 2.0</a>. Built with &#x2665; and Docusaurus.</p>`,
         respectPrefersColorScheme: true,
       },
     } satisfies Preset.ThemeConfig,
+    customFields: {
+      NODE_ENV: process.env.NODE_ENV,
```
matevz, line 248: I'm not sure if we define the build type anywhere. Is it `production` by default when you do `docusaurus build`?

matevz (approved):

docusaurus.config.ts
lubej: It does. It comes from node environment.

lubej:
Closes https://github.com/oasisprotocol/docs/issues/1130

Issue #1310, opened 2025-07-03
[UX/UI]:  Oasis Docs Homepage design and implementation


kaja-osojnik:
### IMPACT

- [ ] High
- [ ] Medium
- [ ] Low

### URGENCY

- [ ] Critical
- [ ] Major
- [ ] Medium
- [ ] Minor

### DESCRIPTION OF ISSUE

Create a new more colourful marketable design for Oasis Docs homepage by using the new design system.

<img width="726" height="855" alt="Image" src="https://github.com/user-attachments/assets/1d7a962d-bea4-4a94-a555-32c9c9f57e64" />

### DEVICE / VIEWPORT

- [x] Desktop
- [x] Mobile

kaja-osojnik:
Implemented in PR [#1324](https://github.com/oasisprotocol/docs/pull/1324)

PR #1324, opened 2025-07-11
Homepage redesign

kaja-osojnik:
Implementation of the new homepage design. Implementation of [1310](https://github.com/oasisprotocol/docs/issues/1310)
Before:
Screenshot 2025-07-11 at 11 33 11

After:
Screenshot 2025-07-11 at 11 34 57


docusaurus.config.ts
```diff
@@ -77,6 +77,12 @@ const config: Config = {
       redirectsOptions,
     ],
   ],
+  stylesheets: [
+    {
+      href: 'https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap',
```
matevz: This is downloaded at compile-time right?

docusaurus.config.ts
```diff
@@ -113,6 +119,7 @@ const config: Config = {
           },
           {
             label: 'Get Involved',
+            description: 'Welcome! This documentation is designed to provide you with a comprehensive understanding and step-by-step guidance for becoming a node operator within the Oasis Network.',
```
matevz: These should go into each .md file at the top, where there is the metadata section. See https://docusaurus.io/docs/markdown-features#front-matter

src/css/custom.css
```diff
@@ -90,7 +92,7 @@ html[data-theme='dark'] .docusaurus-highlight-code-line {
   width: 24px;
   height: 24px;
   display: flex;
-  background: url("data:image/svg+xml,%3Csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='none' stroke='#000' stroke-width='2' d='M12 23c6.075 0 11-4.925 11-11S18.075 1 12 1 1 5.925 1 12s4.925 11 11 11zm0 0c3 0 4-5 4-11S15 1 12 1 8 6 8 12s1 11 4 11zM2 16h20M2 8h20'%3E%3C/path%3E%3C/svg%3E")
+  background: url("data:image/svg+xml,%3Csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='none' stroke='%23000' stroke-width='2' d='M12 23c6.075 0 11-4.925 11-11S18.075 1 12 1 1 5.925 1 12s4.925 11 11 11zm0 0c3 0 4-5 4-11S15 1 12 1 8 6 8 12s1 11 4 11zM2 16h20M2 8h20' /%3E%3C/svg%3E")
```
matevz, line 100: Seems like some URL-string encoding format changes. This needs to be examined and tested. Or did you deliberately change this?

src/css/custom.css
```diff
@@ -100,8 +102,8 @@ html[data-theme='dark'] .docusaurus-highlight-code-line {
 }
 
 [data-theme='dark'] .header-www-link::before {
-  background: url("data:image/svg+xml,%3Csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='none' stroke='#fff' stroke-width='2' d='M12 23c6.075 0 11-4.925 11-11S18.075 1 12 1 1 5.925 1 12s4.925 11 11 11zm0 0c3 0 4-5 4-11S15 1 12 1 8 6 8 12s1 11 4 11zM2 16h20M2 8h20'%3E%3C/path%3E%3C/svg%3E")
-  no-repeat;
+  background: url("data:image/svg+xml,%3Csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='none' stroke='%23fff' stroke-width='2' d='M12 23c6.075 0 11-4.925 11-11S18.075 1 12 1 1 5.925 1 12s4.925 11 11 11zm0 0c3 0 4-5 4-11S15 1 12 1 8 6 8 12s1 11 4 11zM2 16h20M2 8h20' /%3E%3C/svg%3E")
```
matevz, line 110: Same here.

docs/README.mdx
```diff
@@ -1,93 +1,95 @@
...
+  svg={<GetInvolvedSvg className="animatedSvg" />}
+/>
+
+<CustomDocCardSection
+  title="Build on Oasis"
+  description={
+    <>
+      Contains learning material for the smart contract developers. Since the
+      Oasis platform is best known for confidentiality, the most notable
+      ParaTime is <a href="/build/sapphire/">Oasis Sapphire</a>, an{" "}
```
matevz: These links do not seem to be clickable in HTML.

src/theme/DocCard/index.tsx
matevz, line 1: The new DocCard component looks dangerous for future Docusaurus updates.
1. Is the new component really needed? Is it rendered inside a CustomDocCardSection or a replacement for every doc card? Should we rather name it `CustomDocCard` or `DocCardSection`?
2. Could we simply swizzle the DocCard https://docusaurus.io/docs/swizzling ?

src/css/custom.css
```diff
@@ -156,3 +158,32 @@ html[data-theme="dark"] {
   padding: 1rem;
   border-left: 3px solid var(--ifm-tabs-color-active-border);
 }
+
+/* Homepage specific styling */
```
matevz, line 167: I think we can also hide the "Getting Started" title on the landing page to save vertical space.

docusaurus.config.ts
kaja-osojnik: Stylesheet is injected at build time, but the actual font files are downloaded at runtime by the browser from the Google Fonts CDN - @matevz should we self-host the fonts instead?

src/css/custom.css
kaja-osojnik: I changed it because the icon wasn't showing in the navigation and there was an error - "error on line 1 at column 87: AttValue: ' expected" when opening the image link.

src/theme/DocCard/index.tsx
kaja-osojnik: 1. CustomDocCardSection is the whole card including Title, description, DocCards (which is just the 'link' section), but happy to rename it for clarity.
2. The DocCard/index.tsx is already the swizzled version.

docs/README.mdx
matevz: Also, can you make the title (optinally) clickable with the link to the corresponding part?

Perhaps add an optional attribute `path` , so

```
<CustomDocCardSection
  title="Build on Oasis"
  path="/build/"
...
```

docusaurus.config.ts
matevz: > Stylesheet is injected at build time, but the actual font files are downloaded at runtime by the browser from the Google Fonts CDN - @matevz should we self-host the fonts instead?

Not sure actually. We should use the same approach as we do in the rest of our products so it's consistent.

src/css/custom.css
matevz: Nice catch.

docusaurus.config.ts
kaja-osojnik: Changed to add via @fontsource-variable package

src/css/custom.css
kaja-osojnik: Removed.

docs/README.mdx
```diff
@@ -1,93 +1,100 @@
...
+/>
+
+<CustomDocCardSection
+  title="Develop Core"
+  path="/core/"
+  description="Whether you want to contribute your code to the core components of the Oasis Network or just learn more about the Oasis consensus layer and other core components, this is the part for you."
+  itemPaths={[
+    "/core/",
+  ]}
+  secondDescription={`Additions or changes to the interoperable Oasis network components are always made with consensus. Similar to the Ethereum's ERC/EIP mechanism Oasis follows formal Architectural Decision Records (ADRs) which are first proposed, voted on and finally implemented, if accepted.`}
```
matevz: Let's simply put this inside the existing `description` attribute as a second paragraph.

docs/README.mdx
```diff
@@ -1,93 +1,100 @@
...
+  secondItemPaths={[
+    "/adrs",
```
matevz, line 101: Move inside `itemPaths`

docusaurus.config.ts
kaja-osojnik: I looked into this — turns out DocCard doesn’t have access to the frontmatter by default, so adding description there doesn’t work out of the box.
we’d need to swizzle and manually wire up the doc metadata, so i’d suggest keeping it as is for now — happy to revisit if we want to go deeper later.

docs/README.mdx
```diff
@@ -1,93 +1,100 @@
...
+<DocCardSection
+  title="Build on Oasis"
+  path="/build/"
+  description={
+    <>
+      Contains learning material for the smart contract developers. Since the
+      Oasis platform is best known for confidentiality, the most notable
+      ParaTime is <a href="/build/sapphire/">Oasis Sapphire</a>, an{" "}
+      EVM-compatible ParaTime with built-in contract state encryption. The Oasis
+      team also prepared a set of libraries called the{" "}
```
matevz: Leftover? Also above and below.

```suggestion
      team also prepared a set of libraries called the
```

external/oasis-sdk
matevz, line 1: Should not be part of this PR.

docusaurus.config.ts
matevz: Let's move the description to the sidebars files then. Links here are just for the footer.

matevz (approved):

PR #1363, opened 2025-07-28
docs: Fix state sync documentation

martintomazic:
Document how to set correct trusted height and hash for the state sync so that they point to sufficiently old block header, to ensure there are available checkpoints.

### DIFF
--- a/docs/node/run-your-node/advanced/sync-node-using-state-sync.md
+++ b/docs/node/run-your-node/advanced/sync-node-using-state-sync.md
@@ -172,8 +172,8 @@ the values you need are `latest_height` and `latest_hash`.
 #### Public Rosetta Gateway
 
 First obtain the network's Genesis document's hash (e.g. from the Networks Parameters Page):
-- mainnet: [bb3d748def55bdfb797a2ac53ee6ee141e54cd2ab2dc2375f4a0703a178e6e55](https://docs.oasis.io/node/mainnet/)
-- testnet: [0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76](https://docs.oasis.io/node/testnet/)
+- [Mainnet](https://docs.oasis.io/node/mainnet/): `bb3d748def55bdfb797a2ac53ee6ee141e54cd2ab2dc2375f4a0703a178e6e55`
+- [Testnet](https://docs.oasis.io/node/testnet/): `0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76`
 
 Query our public Rosetta Gateway instance and obtain the latest height by
 replacing the `<chain-context>` with the value obtained in the previous step:
--- a/docs/node/run-your-node/advanced/sync-node-using-state-sync.md
+++ b/docs/node/run-your-node/advanced/sync-node-using-state-sync.md
@@ -23,6 +23,15 @@ to be able to [detect and submit evidence for a light client attack].
 
 :::
 
+:::tip
+
+After a successful state sync it is always recommended to check if you see
+the same chain as other nodes. This can be done by comparing block hash at
+a recent height with sources that you trust: e.g. your own nodes, trusted nodes
+from external entity, block explorers, etc.
+
+:::
+
 To configure your node to use the state sync, amend your node's configuration
 (i.e. `config.yml`) with (non-relevant fields omitted):
 
@@ -37,9 +46,10 @@ consensus:
   # Enable consensus state sync (i.e. CometBFT light client sync).
   state_sync:
     enabled: true
-  # Configure trusted height & hash for the light client.
+  # Configure trusted period, height and hash for the light client.
   light_client:
     trust:
+      period: {{ trusted_period }}
       height: {{ trusted_height }}
       hash: "{{ trusted_height_hash }}"
 
@@ -49,6 +59,7 @@ consensus:
 
 and replace the following variables in the configuration snippet:
 
+* `{{ trusted_period }}`: Trusted period is the duration for which trust remains valid.
 * `{{ trusted_height }}`: Trusted height defines the height at which your node should trust the chain.
 * `{{ trusted_height_hash }}`: Trusted height hash defines the hash of the block header corresponding to the trusted height.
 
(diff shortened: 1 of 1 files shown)


docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -153,15 +181,44 @@ This will give you output like the following (non-relevant fields omitted):
...
+			"hash": "3076ae195cfeda09ad49a6c74f6f655bc623e526184f814a842b224bf1846223"
 		},
     ...
 	}
 }
 ```
 
-The values you need are `index` and `hash`.
+Assuming blocks happen every 6 seconds, subtract around `140_000` blocks to
+get the height that is around 10 days old and query again:
```
martintomazic, line 210: Ideally we would do the same for the oasis CLI documentation. The problem is CLI does not have show block (at least for consensus).  Once this https://github.com/oasisprotocol/oasis-core/pull/6265 will be released, there will be no more need for the 2 step process, but for now we need to fix this.

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -81,12 +89,32 @@ something like the following in your node's logs:
...
+:::caution
+
+Checkpoints happen approximately once per week. It is important to set
+sufficiently old trusted height and hash, so that the network has at least one
+checkpoint available to serve. Moreover, to prevent a long range attack
+it is recommended that the light client trust period is lower than the unbonding
+period.
+
+We recommend using the default trust period (2 weeks, no need to configure) and
+configuring the trusted height and hash for the block that is around 10 days old.
```
martintomazic: We could omit this altogether. There is also [existing discussion](https://oasisprotocol.slack.com/archives/C075WCGDNJ0/p1752569369335169) about lowering the defaults trust period to e.g. 10 days.

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
martintomazic: Possibly we could add a method that shows consensus block information for a given height. There seems to be similar method for paratimes...

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -2,12 +2,11 @@
...
 The State Sync is a way to **quickly bootstrap** a **full Oasis node** (either a
 [validator node](../validator-node.mdx) or a
-[non-validator node](../non-validator-node.mdx)) by using the
-[CometBFT's Light Client protocol]. It allows one to initialize a node from a
-trusted height, its corresponding block's header and a trusted validator set
-(given in the [genesis document](../../genesis-doc.md)). It does so by securely
-updating the node's trusted state by requesting and verifying a minimal set of
-data from the network's full nodes.
+[non-validator node](../non-validator-node.mdx)) by initializing it from the
+trusted block's header, configured by the trusted height and hash. The node's
```
matevz: "configured" doesn't sound right to me :shrug:

```suggestion
trusted block's header, identified by the trusted height and hash. The node's
```

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -2,12 +2,11 @@
...
+trusted state is securely updated by requesting and verifying a minimal set of
```
matevz:
```suggestion
trusted state is then securely updated by requesting and verifying a minimal set of
```

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -24,6 +23,15 @@ to be able to [detect and submit evidence for a light client attack].
 
 :::
 
+:::tip
+
+After a successful state sync it is always recommended to manually check if you
+see the same chain as others. This can be done by comparing block hash at
+a recent height with sources that you trust: e.g. your own nodes, trusted nodes
+from external entity, block explorers, etc.
```
matevz:
```suggestion
After a successful state sync it is always recommended to check if you
see the same chain as other nodes. This can be done by comparing the block hash at
a recent height with sources that you trust: e.g. your own nodes, trusted nodes
from external entity, block explorers, etc.
```

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -81,12 +89,32 @@ something like the following in your node's logs:
...
 
+:::caution
+
+Checkpoints happen approximately once per week. It is important to set
+sufficiently old trusted height and hash, so that the network has at least one
+checkpoint available to serve. Moreover, to prevent a long range attack
+it is recommended that the light client trust period is lower than the unbonding
+period.
+
+We recommend using the default trust period (2 weeks, no need to configure) and
```
matevz: Debonding period is expressed in epochs, so I would stick to those.

```suggestion
We recommend using the default trust period (currently 336 epochs or ~14 days, no need to configure) and
```

martintomazic: Hm, but this is trust and not debonding period so probably I should swap the order... Also I was wrong, the default one is 30 days. 

Given the trust period importance, I would suggest that docs encourage to set it manually instead of relying on the defaults. I propose setting 12 days for the trust period?

This gives us 2 day difference between trust and debonding period which should be enough time for the worst case: light client sync + submit evidence. Finally, it enables using trusted headers that are around ~10days old which guarantees one available checkpoint height.

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -81,12 +89,32 @@ something like the following in your node's logs:
 To obtain the trusted height and the corresponding block header's hash, use one
 of the following options.
 
+:::caution
+
+Checkpoints happen approximately once per week. It is important to set
+sufficiently old trusted height and hash, so that the network has at least one
+checkpoint available to serve. Moreover, to prevent a long range attack
+it is recommended that the light client trust period is lower than the unbonding
```
matevz:
```suggestion
it is recommended that the light client trust period is lower than the debonding
```

matevz (approved):
I think this can get merged and then we'll do followups. btw. Does ParaTime state sync work the same way?

martintomazic:
>I think this can get merged and then we'll do followups. btw. Does ParaTime state sync work the same way?

Sorry just saw this. I have added fixup commit explaining how to set correct trust ~header~period. Happy to remove also... :)

martintomazic:
> Does ParaTime state sync work the same way?

It is simpler as you use reindexed history (roothash service). In a way consensus is light client for the runtimes, so you don't have this issues with trust period and sufficiently old checkpoints as for every consensus checkpoint checkpointer also creates runtime checkpoint.

The only think you can configure is disabling checkpoint sync and sync via diffs instead from the genesis I think.

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -76,17 +86,54 @@ something like the following in your node's logs:
...
+penalized for their misbehaviour and you may be tricked into following the wrong chain.
+
+:::
+
+We recommend using `trust_period=288h` (12 days). This way the time required
+to verify headers, submit possible misbehavior evidence and penalize nodes
+is still less than the debonding period, giving nodes strong incentive not to lie.
+
+
+
```
matevz:
```suggestion
```

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
```diff
@@ -130,8 +177,8 @@ First obtain the network's Genesis document's hash (e.g. from the Networks Param
 - mainnet: [bb3d748def55bdfb797a2ac53ee6ee141e54cd2ab2dc2375f4a0703a178e6e55](https://docs.oasis.io/node/mainnet/)
 - testnet: [0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76](https://docs.oasis.io/node/testnet/)
```
matevz:
```suggestion
- [Mainnet]: `bb3d748def55bdfb797a2ac53ee6ee141e54cd2ab2dc2375f4a0703a178e6e55`
- [Testnet]: `0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76`

[Mainnet]: https://docs.oasis.io/node/mainnet/
[Testnet]: https://docs.oasis.io/node/testnet/
```

matevz (approved):

docs/node/run-your-node/advanced/sync-node-using-state-sync.md
matevz: I think it's easier for copy&pasting if the hash is non-clickable.

martintomazic: Btw this syntax doesn't work since instead of visible element, markdown interprets it as another link definition, giving you empty bullet.

Instead you need to:
```markdown
 [Mainnet] [Mainnet]: `bb3d748def55bdfb797a2ac53ee6ee141e54cd2ab2dc2375f4a0703a178e6e55`
 
 [Mainnet]: https://docs.oasis.io/node/mainnet/
```

I have kept it inline for now, just inversed what is clickable since simpler :)

PR #1375, opened 2025-08-04
Shorten DocCard Titles

kaja-osojnik:
Make DocCard Titles shorter, to not overlap with the arrow in case of long titles.

Before:
Screenshot 2025-08-04 at 10 52 26

After:
Screenshot 2025-08-04 at 10 51 48


matevz (approved):
Works as expected, thanks.

PR #1437, opened 2025-09-09
docs: Add Ledger error on fresh Linux to FAQ

matevz:
Bumped on this when plugging in Ledger for the first time on a fresh Ubuntu installation.

### DIFF
--- a/docs/general/manage-tokens/faq.mdx
+++ b/docs/general/manage-tokens/faq.mdx
@@ -462,6 +462,64 @@ exchange support and provide them the link to your account on the
 [Oasis Scan](https://www.oasisscan.com) where they can verify all transactions.
 To learn more about this issue, read the [Manage tokens](README.mdx) chapter.
 
+### Error when accessing Ledger on Linux
+
+The following error is common on fresh Linux installations when accessing your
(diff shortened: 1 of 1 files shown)


docs/general/manage-tokens/faq.mdx
```diff
@@ -462,6 +462,23 @@ exchange support and provide them the link to your account on the
...
+```
+Error: ledger: failed to connect to device: hidapi: failed to open device
+```
+
+If you are running the application as normal user, you need to add some udev
+rules to fix the USB device permissions. The Ledger team prepared a script
+that does it for you:
+
+```shell
+wget -q -O - https://raw.githubusercontent.com/LedgerHQ/udev-rules/master/add_udev_rules.sh | sudo bash
```
kostko: Running things from the internet via `sudo` seems quite dangerous. The script is minimal, maybe just reproduce it here or use a stable reference or something? There is also the `ledger-wallets-udev` package on Ubuntu which maybe installs the same rules?

matevz: I examined and tested both the Ubuntu package and adapted the upstream script to work by copy&pasting.

kostko (approved):

PR #1472, opened 2025-10-01
docs/node/run-your-node: Document use of client-stateless mode

anzoman:
Oasis Core [25.5](https://github.com/oasisprotocol/oasis-core/releases/tag/v25.5) has brought support for stateless client nodes, which are particularly suitable for ROFL nodes. 
We are documenting that here.

### DIFF
--- a/docs/node/run-your-node/paratime-client-node.mdx
+++ b/docs/node/run-your-node/paratime-client-node.mdx
@@ -76,23 +76,33 @@ accessible on the internet (for `TCP` and `UDP` traffic).
 [Mainnet]: ../network/mainnet.md
 [Testnet]: ../network/testnet.md
 
-### Steteless Client Node (optional)
+### Stateless Client Node (optional)
 
-Client node requires sufficient disk space and must wait for synchronization.
-A stateless client node fetches state via configured gRPC endpoints (provider node addresses) and uses a light client
-to verify data against state roots from block headers. It can be started using the following configuration:
+Client node requires sufficient disk space and must wait for state synchronization.
+A stateless client node avoids these problems by fetching the state and light blocks via configured gRPC endpoints
+(provider node addresses) and using a light client to verify the data against state roots from block headers.
+It can be started using the following configuration:
 
 ```yaml
 mode: client-stateless
 # ... sections not relevant are omitted ...
 consensus:
+    # Light client configuration
+    light_client:
+        trust:
+            # ... sections less relevant are omitted ...
     providers:
         - <node-address-1>
         - <node-address-2>
         # Add more node addresses as needed
 ```
 
-To ensure compatibility, all provider nodes specified must be running the latest version of Oasis Core.
+See the [State Sync](./advanced/sync-node-using-state-sync.md) documentation for configuring trusted period, height and
+hash for the light client.
+
+To ensure compatibility, all provider nodes specified must be running the latest version of Oasis Core. The provider
+address contains the `oasis1` prefix. It can also be an IP of a node on the network, or a path to the socket of a local
+node.
 
 ### TEE ParaTime Client Node (optional)
 
--- a/docs/node/run-your-node/paratime-client-node.mdx
+++ b/docs/node/run-your-node/paratime-client-node.mdx
@@ -78,6 +78,12 @@ accessible on the internet (for `TCP` and `UDP` traffic).
 
 ### Stateless Client Node (optional)
 
+:::info
+
+Stateless client is still in early stage, and thus not suitable for use cases where you need high availability.
+
+:::
+
 Client node requires sufficient disk space and must wait for state synchronization.
 A stateless client node avoids these problems by fetching the state and light blocks via configured gRPC endpoints
 (provider node addresses) and using a light client to verify the data against state roots from block headers.
@@ -97,12 +103,30 @@ consensus:
         # Add more node addresses as needed
 ```
 
-See the [State Sync](./advanced/sync-node-using-state-sync.md) documentation for configuring trusted period, height and
-hash for the light client.
+We recommend configuring a recent trust root (e.g. 1000 blocks old), that is younger than your providers last retained
+height, taking their pruning into account.
 
-To ensure compatibility, all provider nodes specified must be running the latest version of Oasis Core. The provider
-address contains the `oasis1` prefix. It can also be an IP of a node on the network, or a path to the socket of a local
-node.
+:::info
+
+See the [State Sync](./advanced/sync-node-using-state-sync.md) documentation for more info on configuring trusted
+period, height and hash for the light client.
+
+:::
+
+The provider address can be a domain name, IP of a node on the network, or a path to the socket of a local node.
+To ensure compatibility, all provider nodes specified must be running the latest version of Oasis Core.
+
+:::note
+
+Make sure you allow all the methods listed in [this example](../grpc.mdx#envoy) in your gRPC proxy for Oasis node.
+
+:::
+
+:::tip
+
+You can use `grpc.oasis.io:443` or `grpc.testnet.oasis.io:443` as provider addresses for Mainnet and Testnet.
+
+:::
 
 ### TEE ParaTime Client Node (optional)
 
--- a/docs/node/run-your-node/paratime-client-node.mdx
+++ b/docs/node/run-your-node/paratime-client-node.mdx
@@ -76,6 +76,24 @@ accessible on the internet (for `TCP` and `UDP` traffic).
 [Mainnet]: ../network/mainnet.md
 [Testnet]: ../network/testnet.md
 
+### Steteless Client Node (optional)
+
+Client node requires sufficient disk space and must wait for synchronization.
+A stateless client node fetches state via configured gRPC endpoints (provider node addresses) and uses a light client
+to verify data against state roots from block headers. It can be started using the following configuration:
+
+```yaml
+mode: client-stateless
+# ... sections not relevant are omitted ...
+consensus:
+    providers:
+        - <node-address-1>
+        - <node-address-2>
+        # Add more node addresses as needed
+```
+
+To ensure compatibility, all provider nodes specified must be running the latest version of Oasis Core.
+
 ### TEE ParaTime Client Node (optional)
 
 If your node requires the ability to issue queries that can access confidential
--- a/docs/node/grpc.mdx
+++ b/docs/node/grpc.mdx
@@ -117,17 +117,28 @@ static_resources:
   Consensus/(\
     EstimateGas|\
     GetBlock|\
+    GetBlockResults|\
     GetChainContext|\
     GetGenesisDocument|\
+    GetLastRetainedHeight|\
+    GetLatestHeight|\
+    GetLightBlock|\
     GetNextBlockState|\
     GetParameters|\
     GetSignerNonce|\
     GetStatus|\
     GetTransactions|\
     GetTransactionsWithResults|\
+    GetTransactionsWithProofs|\
     GetUnconfirmedTransactions|\
     MinGasPrice|\
-    SubmitTx)|\
+    SubmitEvidence|\
+    SubmitTx|\
+    SubmitTxWithProof|\
+    StateSyncGet|\
+    StateSyncGetPrefixes|\
+    StateSyncIterate|\
+    WatchBlocks)|\
   Governance/(\
     ActiveProposals|\
     ConsensusParameters|\
@@ -157,6 +168,7 @@ static_resources:
   RuntimeClient/(\
     CheckTx|\
     GetBlock|\
+    GetBlockResults|\
     GetEvents|\
     GetGenesisBlock|\
     GetLastRetainedBlock|\
--- a/docs/node/run-your-node/README.mdx
+++ b/docs/node/run-your-node/README.mdx
@@ -84,9 +84,14 @@ one or more [ROFL apps].
 ParaTimes such as Sapphire and Cipher that supports confidential smart contact
 queries.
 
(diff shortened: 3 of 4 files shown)


docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,24 @@ accessible on the internet (for `TCP` and `UDP` traffic).
 [Mainnet]: ../network/mainnet.md
 [Testnet]: ../network/testnet.md
 
+### Steteless Client Node (optional)
+
+Client node requires sufficient disk space and must wait for synchronization.
```
gw0: Maybe rephrase it similar as "TEE ParaTime Client Node", e.g.: If you want to decrease startup synchronization time and disk usage, consider running a stateless client node..."

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,24 @@ accessible on the internet (for `TCP` and `UDP` traffic).
 [Mainnet]: ../network/mainnet.md
 [Testnet]: ../network/testnet.md
 
+### Steteless Client Node (optional)
```
peternose: You could add a short description and link under `run-your-node` under client nodes under observer node. 

```
Stateless Node is a special type of client node that operates without maintaining state, allowing it to bootstrap instantly, making it particularly well-suited for ROFL development.
```

docs/node/run-your-node/paratime-client-node.mdx
peternose: Agree. Just don't start paragraph with `If`.

peternose:
```suggestion
Client node requires sufficient disk space and must wait for state synchronization.
```

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,24 @@ accessible on the internet (for `TCP` and `UDP` traffic).
...
+
+Client node requires sufficient disk space and must wait for synchronization.
+A stateless client node fetches state via configured gRPC endpoints (provider node addresses) and uses a light client
```
peternose:
```suggestion
A stateless client node avoids these problems by fetching the state and light blocks via configured gRPC endpoints (provider node addresses) and using a light client to verify the data against state roots from block headers.
```

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,24 @@ accessible on the internet (for `TCP` and `UDP` traffic).
...
+to verify data against state roots from block headers. It can be started using the following configuration:
+
+```yaml
+mode: client-stateless
+# ... sections not relevant are omitted ...
+consensus:
```
peternose, line 95: You also need to define trust root. We already have this somewhere, not sure where 🤔

peternose (commented):
Looks good, plain and simple. We will update this latter once we implement levels of statelessness (consensus yes/no, runtimes yes/no).

docs/node/run-your-node/paratime-client-node.mdx
peternose: @martintomazic Where did you put docs for trust root and suggested trust period?

martintomazic: https://docs.oasis.io/node/run-your-node/advanced/sync-node-using-state-sync ([src](https://github.com/oasisprotocol/docs/blob/main/docs/node/run-your-node/advanced/sync-node-using-state-sync.md))

Fixed some stuff here recently. As promised yesterday, will add a cli command that calculates suggested trust and augment the docs again.

In the long run we could consider moving it out of _"Advance"_...

peternose: We should probably have in the example only simple config, e.g. `consensus: light_client: trust: # visit light client for light client configuration`, and a link to light client config, so that we don't repeat this stuff multiple times.

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,24 @@ accessible on the internet (for `TCP` and `UDP` traffic).
...
+    providers:
+        - <node-address-1>
```
ptrus, line 101: Should probably mention some examples on how to the obtain these address.

peternose: Maybe we could also mention that the address could be an IP of a node on the network, or a path to the socket of a local node.

ptrus: Should we also enable `consensus.WatchBlock` on our public endpoints (e.g. tesntet.grpc.oasis.io?). Currently it's not whitelisted: https://github.com/oasisprotocol/internal-ops/blob/9af8ae8e752e3e26b60df6c96049629fc84f4b54/ansible/roles/oasis-node/defaults/main.yml#L616-L630

Probably no reason not to whitelist it, for example the runtime.WatchBlocks is allowed: https://github.com/oasisprotocol/internal-ops/blob/9af8ae8e752e3e26b60df6c96049629fc84f4b54/ansible/roles/oasis-node/defaults/main.yml#L672

martintomazic: I think this is intentional as producer (server) stores per subscriber data in memory. Meaning if you have many slow subscribers this becomes prone to memory leak... https://github.com/oasisprotocol/oasis-core/issues/6315 is somehow related. Not sure why we don't deem runtime watch blocks "dangerous"

ptrus: Should we update the stateless client mode to not rely on WatchBlocks then? (or have a fallback to polling if its not available)

martintomazic: I think @peternose already had some ideas about polling by implementing `WatchBlocks` via `GetBlocks` internally?

Complementary to that is to start running dedicated "provider" nodes ourselves:  client nodes with #6315 fixed, including a sensible (configurable) drop policy (e.g. max 1h of blocks in memory per subscriber) + sensible  rate limit `WatchBlocks` connections (considering they much more expensive). Given node internally relies o no drop policy until this is fixed we could temporary add a separate method? Brainstorming...

**Update**: The more I think about it polling should easier and faster to implement unless we require some crazy latency here...

ptrus: > Complementary to that is to start running dedicated "provider" nodes ourselves

I don't think we need separate dedicated "provider" nodes. We already run public nodes that _provide_ the data.

> drop policy (e.g. max 1h of blocks in memory per subscriber

No need to worry about this for now. Since our public nodes are behind Cloudflare, it has hard cap connection limit of ~100 seconds. So clients will be dropped and need to reconnect way sooner.

martintomazic: Hm we should check then everything works as expected with disconnections/missing block gaps. I believe initial versions didn't take this into consideration, but maybe this was fixed already. update: or again use poling so this is not an issue at all.

peternose: I wouldn’t optimize prematurely. Let’s expose the `WatchBlocks` method, and if things go south, we can disable/limit it and then optimize.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Updated.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Added.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Refactored.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Tnx, updated.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Sounds good, added a rough light client config part and mentioned state sync docs for more info.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: We'll do that in https://github.com/oasisprotocol/internal-ops/pull/1182.

I have updated the part of the docs about addresses.
Is there a nice way to obtain node addresses that can be providers for stateless clients?

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,34 @@ accessible on the internet (for `TCP` and `UDP` traffic).
...
+        - <node-address-2>
+        # Add more node addresses as needed
+```
+
+See the [State Sync](./advanced/sync-node-using-state-sync.md) documentation for configuring trusted period, height and
+hash for the light client.
+
+To ensure compatibility, all provider nodes specified must be running the latest version of Oasis Core. The provider
+address contains the `oasis1` prefix. It can also be an IP of a node on the network, or a path to the socket of a local
```
ptrus: > The provider address contains the `oasis1` prefix

I don't think this is true?

docs/node/run-your-node/paratime-client-node.mdx
ptrus: > Is there a nice way to obtain node addresses that can be providers for stateless clients?

Could use existing OPF hosted nodes at grpc.oasis.io/testnet.grpc.oasis.io* (are there any other known public nodes run by anyone else?). Otherwise, the user can also use their own nodes.

\* once we update them to expose the WatchBlocks method, and test it to ensure it works.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Ok, I see, will drop that part.

We can also state the (only?) two possible ways of specifying addresses:

- `{{ node_entity_id }}@{{ node_ip_address }}`
- `/node/run/internal.sock`

Does that seem correct?

ptrus: ```{{ node_entity_id }}@{{ node_ip_address }}```

I don't think it's the entity id? I think it's the public key of the TLS certificate?  And it's optional, If included, it specifies the expected peer identity for TLS authentication.
Also doesn't need to be an IP address. Address like `grpc.oasis.io:443` also work.

docs/node/run-your-node/paratime-client-node.mdx
martintomazic: I experimented a bit with stateless client and don't think this is the appropriate reference here as it tells you how to set sufficiently old trust so that consensus checkpoints are available.

With stateless client, there is no need for "sufficiently old" trust since the node is stateless (i.e. no checkpoints will be used).

Further experiments with the paratime client using the stateless client consensus backend showed that taking around 10 days old trust, runtime history reindex takes few hours (5h+). Strangely had to restart in between.  Given that only after reindex is finished, your runtime client worker is ready, and that with stateless client, runtime storage worker is disabled (no checkpoints on the runtime side as well), there is little point in using this configuration.

So my suggestion is to take the latest, or 1000 blocks old trust height for the stateless client node. update: The only reason for non-latest is if you provider is behind... Otherwise as finality with cometbft is instant the latest is fine.

martintomazic: Related [comment](https://github.com/oasisprotocol/oasis-core/pull/6385#issuecomment-3479828582).
>Nevertheless, my suggestion is to reflect this requirement under Setting trust root for stateless client node - "configured trust should be younger than your providers last retained height, taking their pruning and your speed of reindexing into account". See existing discussion https://github.com/oasisprotocol/docs/pull/1472#discussion_r2451969014. Realistically if you configure a recent/latest trust as https://github.com/oasisprotocol/docs/pull/1472#discussion_r2485785109 this is not a problem at all.

Probably we may also want to note that stateless client is still in early stage, and thus not suitable for use cases where you need 99.9%+ uptime/liveness?

anzoman: Nice, these are all very good points, so I added them to the docs.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Yeah, makes sense, updated the text.

docs/node/run-your-node/paratime-client-node.mdx
anzoman: Good point, I mentioned that our public nodes can be used as providers.

docs/node/run-your-node/paratime-client-node.mdx
```diff
@@ -76,6 +76,58 @@ accessible on the internet (for `TCP` and `UDP` traffic).
...
+        trust:
+            # ... sections less relevant are omitted ...
+    providers:
+        - <node-address-1>
+        - <node-address-2>
+        # Add more node addresses as needed
+```
+
+We recommend configuring the most recent trust root, which should be younger than your providers last retained height,
+taking their pruning and your speed of reindexing into account.
```
martintomazic: Sorry was me who suggested this but lets keep this simpler.
```suggestion
We recommend configuring a recent trust root (e.g. 1000 blocks old), that is younger than your providers last retained height,
taking their pruning into account.
```

Or even simpler `We recommend configuring a recent trust root, e.g. one that is 1000 blocks old.`

anzoman: Cool, reworded as you suggested.

ptrus (approved):

PR #1478, opened 2025-10-06
feat: add rofl key genearation use case

### LINKED ISSUE #2298: ROFL Guides / Examples
Starting with:

- [ ] 1. How to sign message on EVM chain e.g. Base

ROFL Guides/Examples Doc: https://docs.google.com/document/d/1fr5-i2c-slIae2XkplsjCr_sA31ELeT9Epq_BbZe_AE/edit?tab=t.0#heading=h.fzov0g9g8u4w

uniyalabhishek:
Closes https://github.com/oasisprotocol/oasis-sdk/issues/2298

[PREVIEW](https://deploy-preview-1478--oasisprotocol-docs.netlify.app/build/use-cases/key-generation)

This PR:
- Creates a new use case guide for key generation using `appd` REST
    -  This guide goes through how to sign message on EVM chain. Base in this case
    - Generate key for Base via appd key REST API (secp…)
    - Send ETH(test) to ROFL key
    - (Deploy Contract via ROFL key)
    - Send ETH(test) back to user via RPC

The example repo can be found here: https://github.com/uniyalabhishek/demo-rofl-keygen

### DIFF
--- a/docs/build/use-cases/key-generation.mdx
+++ b/docs/build/use-cases/key-generation.mdx
@@ -1,0 +1,571 @@
+---
+sidebar_label: Cross-Chain Key Generation
+description: Generate an Ethereum-compatible key inside ROFL via appd and use it to sign and send transactions on Base.
+tags: [ROFL, appd, KMS, EVM]
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Cross-Chain Key Generation on ROFL (EVM / Base)
+
+This chapter shows how to build a tiny TypeScript app that **generates a
+secp256k1 key inside ROFL** using the
+**`@oasisprotocol/rofl-client` TypeScript SDK** (which talks to the
+[appd REST API] under the hood), derives an **EVM address**, **signs**
+messages, **deploys a contract**, and **sends** EIP-1559 transactions on
+**Base Sepolia**. We use a simple **smoke test** that prints to logs.
+
+[appd REST API]: ../rofl/features/appd.md
+
+## Prerequisites
+
+This guide requires:
+
+- **Node.js 20+** and **Docker** (or Podman).
+- **Oasis CLI** and at least **120 TEST** tokens in your wallet
+  (use [Oasis Testnet faucet]).
+- Some Base Sepolia test ETH (use [Base Sepolia faucet]) to test sending ETH.
+
+Check [Quickstart Prerequisites] for setup details.
+
+[Quickstart Prerequisites]: ../rofl/quickstart#prerequisites
+[Oasis Testnet faucet]: https://faucet.testnet.oasis.io
+[Base Sepolia faucet]: https://docs.base.org/base-chain/tools/network-faucets
+
+## Init App
+
+Initialize a new app using the [Oasis CLI]:
+
+```shell
+oasis rofl init rofl-keygen
+cd rofl-keygen
+```
+
+## Create App
+
+Create the app on Testnet (100 TEST deposit):
+
+```shell
+oasis rofl create --network testnet
+```
+
+The CLI prints the **App ID** (e.g., `rofl1...`).
+
+## Init a Hardhat (TypeScript) project
+
+```shell
+npx hardhat init
+```
+
+When prompted, **choose TypeScript** and accept the defaults.
+
+Now add the small runtime deps we use outside of Hardhat:
+
+```shell
+npm i @oasisprotocol/rofl-client ethers dotenv @types/node
+npm i -D tsx
+```
+
+Using Hardhat’s TypeScript template, it already created a `tsconfig.json`.
+Add the following so our app code compiles to `dist/`:
+
+```json
+// tsconfig.json
+{
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist"
+  },
+  "include": ["src"]
+}
+```
+
+## App structure
+
+We'll add a few small TS files and one Solidity contract:
+
+```
+src/
+├── appd.ts               # thin wrapper over @oasisprotocol/rofl-client
+├── evm.ts                # ethers helpers (provider, wallet, tx, deploy)
+├── keys.ts               # tiny helpers (checksum)
+└── scripts/
+    ├── deploy-contract.ts  # generic deploy script for compiled artifacts
+    └── smoke-test.ts       # end-to-end demo (logs)
+contracts/
+└── Counter.sol           # sample contract
+```
+
+### `src/appd.ts` — thin wrapper over the SDK
+
+Use the official client to talk to `appd` (UNIX socket) and keep an explicit
+**local‑dev fallback** when running outside ROFL.
+
+<details>
+  <summary>src/appd.ts</summary>
+
+```ts
+import {existsSync} from 'node:fs';
+import {
+  RoflClient,
+  KeyKind,
+  ROFL_SOCKET_PATH
+} from '@oasisprotocol/rofl-client';
+
+const client = new RoflClient(); // UDS: /run/rofl-appd.sock
+
+export async function getAppId(): Promise<string> {
+  return client.getAppId();
+}
+
+/**
+ * Generates (or deterministically re-derives) a secp256k1 key inside ROFL and
+ * returns it as a 0x-prefixed hex string (for ethers.js Wallet).
+ *
+ * Local development ONLY (outside ROFL): If the socket is missing and you set
+ * ALLOW_LOCAL_DEV=true and LOCAL_DEV_SK=0x<64-hex>, that value is used.
+ */
+export async function getEvmSecretKey(keyId: string): Promise<string> {
+  if (existsSync(ROFL_SOCKET_PATH)) {
+    const hex = await client.generateKey(keyId, KeyKind.SECP256K1);
+    return hex.startsWith('0x') ? hex : `0x${hex}`;
+  }
+  const allow = process.env.ALLOW_LOCAL_DEV === 'true';
+  const pk = process.env.LOCAL_DEV_SK;
+  if (allow && pk && /^0x[0-9a-fA-F]{64}$/.test(pk)) return pk;
+  throw new Error(
+    'rofl-appd socket not found and no LOCAL_DEV_SK provided (dev only).'
+  );
+}
+```
+
+</details>
+
+### `src/evm.ts` — ethers helpers
+
+<details>
+  <summary>src/evm.ts</summary>
+
+```ts
+import {
+  JsonRpcProvider,
+  Wallet,
+  parseEther,
+  type TransactionReceipt,
+  ContractFactory
+} from "ethers";
+
+export function makeProvider(rpcUrl: string, chainId: number) {
+  return new JsonRpcProvider(rpcUrl, chainId);
+}
+
+export function connectWallet(
+  skHex: string,
+  rpcUrl: string,
+  chainId: number
+): Wallet {
+  const w = new Wallet(skHex);
+  return w.connect(makeProvider(rpcUrl, chainId));
+}
+
+export async function signPersonalMessage(wallet: Wallet, msg: string) {
+  return wallet.signMessage(msg);
+}
+
+export async function sendEth(
+  wallet: Wallet,
+  to: string,
+  amountEth: string
+): Promise<TransactionReceipt> {
+  const tx = await wallet.sendTransaction({
+    to,
+    value: parseEther(amountEth)
+  });
+  const receipt = await tx.wait();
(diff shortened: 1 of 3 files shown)


docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
+---
+description: Generate an EVM key inside ROFL via appd and use it to sign and send transactions on Base.
+tags: [ROFL, appd, keys, EVM]
```
matevz: I think `KMS` became a standard term now for managing the keys.

```suggestion
tags: [ROFL, appd, KMS, EVM]
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
...
+[Quickstart Prerequisites]: ../rofl/quickstart#prerequisites
+
+## Init App
+
+Initialize a new app using the [Oasis CLI]:
+
+```shell
+oasis rofl init rofl-keygen
+cd rofl-keygen
+````
```
matevz:
````suggestion
```
````

File: docs/build/use-cases/key-generation.mdx [hunk 3]
```diff
@@ -0,0 +1,576 @@
...
+We'll add four small TS files and one Solidity contract:
+
+```
+src/
+  appd.ts        # talks to appd over /run/rofl-appd.sock
+  evm.ts         # ethers helpers (provider, wallet, tx)
+  keys.ts        # tiny helpers (checksum)
+  server.ts      # HTTP API to drive the demo
+contracts/
+  Counter.sol    # optional sample contract
```
matevz: Let's us the standard `tree` command-like project structure style.

```suggestion
src/
├── appd.ts        # talks to appd over /run/rofl-appd.sock
├── evm.ts         # ethers helpers (provider, wallet, tx)
├── keys.ts        # tiny helpers (checksum)
└── server.ts      # HTTP API to drive the demo
contracts/
└── Counter.sol    # optional sample contract
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
...
+
+## Install minimal deps
+
+We keep dependencies lean:
+
+```shell
+npm init -y
+npm i express ethers zod dotenv
+npm i -D typescript tsx @types/node @types/express hardhat
+npx tsc --init --rootDir src --outDir dist --module NodeNext --target ES2022
```
matevz: I'd just use `hardhat init` and select typescript. Then install any extra deps. See how we did it for Sapphire: https://github.com/oasisprotocol/sapphire-paratime/blob/main/docs/quickstart.mdx#init-a-new-hardhat-project

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
+---
+description: Generate an EVM key inside ROFL via appd and use it to sign and send transactions on Base.
```
matevz: For those who don't know what EVM stands for.

```suggestion
description: Generate an Ethereum-compatible key inside ROFL via appd and use it to sign and send transactions on Base.
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
...
+          resolve(id);
+        });
+      }
+    );
+    req.on("error", reject);
+    req.end();
+  });
+}
+
+export async function getEvmPrivateKey(keyId: string): Promise<string> {
```
matevz: Follow best security practices and use `SecretKey` instead of `PrivateKey`.

```suggestion
export async function getEvmSecretKey(keyId: string): Promise<string> {
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
...
+
+<details>
+  <summary>src/keys.ts</summary>
+
+```ts
+import { Wallet, getAddress } from "ethers";
+
+export function privateKeyToWallet(pkHex: string): Wallet {
+  return new Wallet(pkHex);
+}
```
matevz:
```suggestion
export function secretKeyToWallet(skHex: string): Wallet {
  return new Wallet(skHex);
}
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
...
+
+export function checksumAddress(addr: string): string {
+  return getAddress(addr);
+}
+```
+
+</details>
+
+### `src/server.ts` — minimal HTTP API
```
matevz: I think running another HTTP API is not a good idea. There is a REST server already running and accessible at `/run/rofl/appd.sock` for this purpose.

Users that want to have EVM-signing capability would need to integrate your HTTP API anyway, so why not connecting to appd directly? Instead of server.ts I'd write a simple test that uses appd, get the key, sign an EVM transaction and submit it to Base.

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,576 @@
...
+```
+
+Look for the **Proxy** section (e.g., `https://p8080.mXXX.test-proxy...`).
+
+## End‑to‑end (Base Sepolia)
+
+1. **Get App ID**
+
+   ```shell
+   curl -s https://YOUR-PROXY/app-id | jq
```
matevz: `YOUR_PROXY` is the machine name defined in compose.yaml. Also, https won't work out of the box since the domain is not managed by CA and self-signed certs are not allowed.

matevz (commented):
First pass

docs/build/use-cases/key-generation.mdx
uniyalabhishek: yes, thanks! updated

docs/build/use-cases/key-generation.mdx
uniyalabhishek: removed proxy for the example since it's not needed.

uniyalabhishek:
Also updated example repo: https://github.com/uniyalabhishek/demo-rofl-keygen

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,626 @@
...
+  <summary>src/appd.ts</summary>
+
+```ts
+import { request } from "node:http";
+import { existsSync } from "node:fs";
+
+const APPD_SOCKET = "/run/rofl-appd.sock";
+type KeyKind = "secp256k1" | "ed25519" | "raw-256" | "raw-386";
+
+export async function generateKey(
```
ptrus: Should we create a small rofl-client lib in oasis-sdk (like we have for [python](https://github.com/oasisprotocol/oasis-sdk/tree/main/rofl-client/py)) to reduce the boilerplate?. Or alternatively use python for the example, since that one already has the lib :-) 

Not sure what the policy is in regards to languages used for the examples.

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,626 @@
...
+  }
+  throw new Error("Timed out waiting for funding.");
+}
+
+async function main() {
+  const appId = await getAppId().catch(() => null);
+  console.log(`ROFL App ID: ${appId ?? "(unavailable outside ROFL)"}`);
+
+  const sk = await getEvmSecretKey(KEY_ID);
+  const wallet = secretKeyToWallet(sk).connect(makeProvider(RPC_URL, CHAIN_ID));
```
ptrus, line 280: I think the example should mention that we trust the provider here, and ideally (for production examples) a light-client (e.g., Helios) would be used when fetching state of the remote blockchains, so that the provider responses can be verified.

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,626 @@
...
+that prints to logs.
+
+[appd REST API]: ../rofl/features/appd.md
+
+## Prerequisites
+
+This guide requires:
+
+- **Node.js 20+** and **Docker** (or Podman).
+- **Oasis CLI** and at least **120 TEST** tokens in your wallet.
```
ptrus: Could link to the oasis-faucet here, for the test tokens?

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,626 @@
...
+- A Base Sepolia faucet or funds to test sending ETH.
```
ptrus: Maybe link to some faucet here? (is there a well-known one for Base Sepolia?). Otherwise just mention that you need XX test funds. (like you do in the previous line).

docs/build/use-cases/key-generation.mdx
kostko: Yes, we should add a TypeScript client as well.

uniyalabhishek: yes, makes sense to simplify. will create a TS client

uniyalabhishek: done

sidebarBuild.ts
```diff
@@ -19,6 +19,7 @@ export const sidebarBuild: SidebarsConfig = {
       items: [
         'build/use-cases/price-oracle',
         'build/use-cases/tgbot',
+        'build/use-cases/key-generation',
```
matevz, line 22: Nit: The title is a bit too long for the sidebar. Can you set its sidebar `label` to "Cross-Chain Key Generation".

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,592 @@
...
+      - KEY_ID=${KEY_ID:-evm:base:sepolia}
+      - BASE_RPC_URL=${BASE_RPC_URL:-https://sepolia.base.org}
+      - BASE_CHAIN_ID=${BASE_CHAIN_ID:-84532}
+    volumes:
+      - /run/rofl-appd.sock:/run/rofl-appd.sock
+```
+
+</details>
+
+Add a Dockerfile that builds TS and compiles the contract, runs the
```
matevz, line 438: I'd put the Dockerfile instructions first and then compose below.

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+ENV NODE_ENV=production
+CMD ["sh", "-c", "node dist/scripts/smoke-test.js || true; tail -f /dev/null"]
+```
+
+</details>
+
+## Scripts
+
+Add handy scripts to `package.json`:
```
matevz: These aren't really needed for running ROFL right? I'd omit them then, because they're part of the git repo example anyway for serious projects.

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,592 @@
...
+    "deploy-contract": "node dist/scripts/deploy-contract.js",
+    "compile:contracts": "hardhat compile",
+    "deploy-counter": "npm run compile:contracts && node dist/scripts/deploy-contract.js ./artifacts/contracts/Counter.sol/Counter.json \"[]\""
+  }
+}
+```
+
+## Build the image
+
+ROFL runs on **x86_64/TDX**, so build amd64 images:
```
matevz:
```suggestion
ROFL only runs on Intel TDX-enabled hardware so don't forget to pass the `--platform linux/amd64` parameter if you're compiling images on a different host (e.g. macOS):
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+```shell
+docker buildx build --platform linux/amd64 \
+  -t docker.io/YOURUSER/rofl-keygen:0.1.0 --push .
+```
+
+(Optionally pin the digest and use `image: ...@sha256:...` in compose.)
```
matevz:
```suggestion
For extra security and verifiability pin the digest and use `image: ...@sha256:...` in `compose.yaml`.
```

docs/build/use-cases/key-generation.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+## Build ROFL bundle
+
+Before running the `oasis rofl build` command, make sure to update the
+`services.demo.image` in `compose.yaml` to the image you built.
+
+:::note
+For TypeScript projects, image size may be larger, update the `rofl.yaml`
+`resources` section: `memory`: `1024` & `storage.size`: `4096`.
```
matevz:
```suggestion
`resources` section to at least: `memory: 1024` and `storage.size: 4096`.
```

matevz (approved):
A few nits, otherwise excellent job, thanks!

matevz (commented):
Also, don't forget to add the new chapter to https://github.com/oasisprotocol/docs/blob/main/docs/build/README.mdx#build-on-oasis

Issue #1506, opened 2025-10-23
docs/Run your node: Update hardware requirements


martintomazic:
Update the current hardware requirements, specifically [storage](https://docs.oasis.io/node/run-your-node/prerequisites/hardware-recommendations#storage). 

Currently, they portray impression that running a Consensus + Sapphire node (Validator + Compute node) requires 2TB of the disk space. This are hardware requirements for nodes that are able to serve all historical state... Serving historical state is NOT validators responsibility.

Instead they should have pruning set to few weeks max - see https://github.com/oasisprotocol/docs/pull/1526. This makes them fast, enables checkpoint creation that is quick and has little IO overhead,  resulting in a much smaller disk footprint, and thus cost. E.g. 100GB should be enough.

Finally, we should organize _Run your node_ sections from most to least accessible options.  E.g. hardware requirements should reflect this order: 
1. Stateless
2. Non-Validator/Paratime-Client (with pruning)
3. Validator/Executor (with pruning)
4. Observer (with pruning)
5. Non-Validator/Paratime-Client/Observer with all historical state.
6. Keymanager/Sentry/Archive

Partially overlaps with #1469.

amela:
Thx for the ideas.

martintomazic:
>Finally, we should organize Run your node sections from most to least accessible options. E.g. hardware requirements should reflect this order:

If we decide for such reorg, it would be nice to avoid the pattern of the partial configurations, that reference 3 other chapters, in order to augment the config. I understand we don't want to repeat snippets, but this unfortunately results in the very bad UX for the end user.

_Depending on the role/use case_ (that should have a dedicated section), _I want to directly copy the whole config_. The _only thing_ I expect to _edit_ are _`$VAR_NAME`_ in the config. For those we could have how to guides (e.g. pruning, trust root etc...), that could be further split per role. E.g. setting trust root for stateless client is different then setting it for the paratime client node.

If we decide to proceed with this, "the new organization" is actually a separate issue than the "update hardware requirements". Just idea, priority low, won't be working on this anytime soon. :)

PR #1526, opened 2025-11-04
docs/node/run-your-node/advanced: Add new pruning section

### LINKED ISSUE #1469: docs/Run your node: Add dedicated section about pruning
We should add a dedicated section about pruning, probably under `Maintenance` section?

1. How to configure it for consensus and paratime nodes.
    * We have this currently under hardware requirements.
2. Suggested values for the pruning.
3. How to guarantee optimal disk usage:
    * For now by pruning from the start.
    * Later we will add the `oasis-node storage compact` command there, possibly "offline pruning".
4. Explain that with many versions, pruning is quite slow (especially for the runtime state) and that operators should consider (temporary) increasing BadgerDB block cache size.
    * Especially when keeping months of the data, if possible. Pending some more experiments from oasisprotocol/oasis-core#6334.
5. How much extra free disk space operator should have available to guarantee compaction success.

martintomazic:
Closes #1469.

## Considerations
* We should be careful to not trigger data availability issues of the whole network, with new pruning suggestions.
* Encouraging validators to only keep minimal state may allow us to also encourage them to enable checkpoints for consensus and runtime state (possibly making this default), solving another availability issue :).

## Possible follow-up
 https://github.com/oasisprotocol/docs/issues/1506
 
 ## Hardening data availability.
*  https://github.com/oasisprotocol/oasis-core/issues/6422
* https://github.com/oasisprotocol/oasis-core/issues/6423

### DIFF
--- a/docs/node/run-your-node/advanced/pruning.md
+++ b/docs/node/run-your-node/advanced/pruning.md
@@ -1,0 +1,78 @@
+# Pruning State
+
+To reduce hardware requirements and improve the overall latency of the node,
+operators are encouraged to prune their state (unless also serving historical
+state).
+
+As will be explained in the following sections, pruning should be configured from
+the start, i.e. late changes to the pruning configuration may not be optimal for
+the node's health.
+
+## Consensus Pruning
+
+To configure pruning of the consensus data amend your node's configuration with:
+
+```yaml
+# ... sections not relevant are omitted ...
+consensus:
+    # ... sections not relevant are omitted ...
+    prune:
+        strategy: "keep_n"
+        num_kept:  <n>
+```
+
+where `<n>` is the number of blocks and state versions that will be kept.
+
+### Suggested Pruning Configuration
+
+For normal node operation, the minimum requirements is `n=250_000`. However,
+it is recommended to set it to `n=400_000`. Assuming blocks are produced every 6
+seconds, the latter corresponds to preserving 1 month of the consensus data.
+
+Operators that are not resource-restrained and want to contribute to the overall
+network health are encouraged to set `n=5_000_000` (approximately 1 year of data).
+
+### Late Pruning
+
+Due to the [LSM Tree][] design of the underlying databases, enabling pruning after
+your node has been running for a while, or changing configuration to retain less data,
+may not work as expected.
+
+To change pruning configuration operators must follow this steps:
+
+1. Gracefuly shutdown the node (validators should take [special care][graceful shutdown]).
+2. Configure pruning as described in the [Consensus Pruning](#consensus-pruning) section above.
+3. Run offline [pruning command].
+4. Run the [compaction command].
+5. Start the node again.
+
+[LSM tree]: https://en.wikipedia.org/wiki/Log-structured_merge-tree
+[graceful shutdown]: ../maintenance/shutting-down-a-node.md
+[pruning command]: https://github.com/oasisprotocol/oasis-core/blob/master/docs/oasis-node/cli.md#prune-experimental
+[compaction command]: https://github.com/oasisprotocol/oasis-core/blob/master/docs/oasis-node/cli.md#compact-experimental
+
+## ParaTime Pruning
+
+To configure pruning of the ParaTime data amend your node's configuration with:
+
+```yaml
+# ... sections not relevant are omitted ...
+runtime:
+    # ... sections not relevant are omitted ...
+    prune:
+        strategy: "keep_last"
+        num_kept:  <n>
+```
+
+where `<n>` is the number of ParaTime's state versions that will be kept.
+
+### Suggested Pruning Configuration
+
+ParaTime state is much larger than the consensus state. As a consequence, as the
+number of versions in the state database grows, pruning becomes slower and slower.
+For this reason **ParaTime state pruning must be configured
+from the start or not configured at all**.
+
+To configure ParaTime pruning it is recommended to set `n=250_000`. The maximum value is
+`n=400_000`. If you need to preserve more data (e.g. nodes serving historical
+state) you will have to keep the entire ParaTime's state from the genesis.
--- a/sidebarNode.ts
+++ b/sidebarNode.ts
@@ -106,6 +106,7 @@ export const sidebarNode: SidebarsConfig = {
           },
           items: [
             'node/run-your-node/advanced/sync-node-using-state-sync',
+            'node/run-your-node/advanced/pruning',
             'node/run-your-node/advanced/copy-state-from-one-node-to-the-other',
             'node/run-your-node/advanced/remote-signer',
           ],
@@ -165,7 +166,7 @@ export const sidebarNode: SidebarsConfig = {
               label: 'Mainnet',
             },
             {
-              type: 'doc', 
+              type: 'doc',
               id: 'node/reference/upgrade-logs/testnet',
               label: 'Testnet',
             },
--- a/docs/node/run-your-node/prerequisites/hardware-recommendations.md
+++ b/docs/node/run-your-node/prerequisites/hardware-recommendations.md
@@ -152,18 +152,12 @@ running a separate archive node.
 
 :::info
 
-You can configure your node _not to_ keep a complete state from the genesis
-onward. This will reduce the amount of storage required for the consensus and
-ParaTime state.
+Unless running a node to preserve historical state you are encouraged _not to_
+keep a complete state from the genesis onward. This will significantly reduce
+the amount of storage required for the consensus and ParaTime state and improve
+the overall latency of the storage related operations.
 
-To enable pruning of the consensus state set the
-`consensus.prune.strategy` and
-`consensus.prune.num_kept` parameters appropriately in your
-[node's configuration].
-
-To enable pruning of the ParaTime state set the
-`runtime.prune.strategy` and `runtime.prune.num_kept`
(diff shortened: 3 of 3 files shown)


docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,77 @@
...
+### Suggested Pruning Configuration
+
+Paratime state is much larger than the consensus state. As a consequence, as the
+number of versions in the state database grows, pruning becomes slower and slower.
+ For this reason **paratime state pruning must be configured
+from the start or not configured at all**.
+
+To configure pruning it is recommended to set `n=250_000`. The maximum value is
+`n=400_000`. If you need to preserve more data (e.g. nodes serving historical
+state) you will have to keep the entire state from the genesis.
```
martintomazic: Unfortunately if you preserve runtime state, to serve historical queries, you are also forced to preserve all the consensus data, as consensus light block are computed ad-hoc ([see](https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/client/committee/node.go#L169:L190)). A requirement that should probably be documented somewhere.

We could open an issue, so that consensus light blocks are not computed ad-hoc and are instead stored in the separate DB, with different pruning config. This would allow to run Observer node with much lower hardware requirements, assuming most of the consensus data could be dropped.

Not sure this is a critical priority right now, ~but issue definitely cannot hurt...~: https://github.com/oasisprotocol/oasis-core/issues/6430

sidebarNode.ts
```diff
@@ -106,6 +106,7 @@ export const sidebarNode: SidebarsConfig = {
           },
           items: [
             'node/run-your-node/advanced/sync-node-using-state-sync',
+            'node/run-your-node/advanced/pruning',
```
martintomazic, line 109: There is nothing "advanced" about pruning, in fact pruning should probably enabled via sane defaults (so possibly no need to even configure it)....

When documentation is reorganized (out of scope of this PR), this could probably be "How to guide"

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,77 @@
...
+### Late Pruning
+
+Due to the [LSM Tree][] design of the underlying databases, enabling pruning after
+your node has been runing for a while, or possibly changing configuration to retain
+less data, may may not work as expected.
+
+To change pruning configuration operators must follow this steps:
+1. Gracefuly shutdown the node (validators should take [special care][graceful shutdown]).
+2. Configure pruning as described in [Consensus Pruning](#consensus-pruning).
+3. Run offline pruning (TODO insert link).
```
martintomazic: ~We can ad whole `Late Pruning` section as a follow-up once [offline pruning command](https://github.com/oasisprotocol/oasis-core/pull/6341) is merged.~

update: Pending `25.9` release and documentation bump.

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Suggested Pruning Configuration
+
+Paratime state is much larger than the consensus state. As a consequence, as the
+number of versions in the state database grows, pruning becomes slower and slower.
+ For this reason **paratime state pruning must be configured
+from the start or not configured at all**.
+
+To configure pruning it is recommended to set `n=250_000`. The maximum value is
+`n=400_000`. If you need to preserve more data (e.g. nodes serving historical
```
martintomazic: We should preserve enough data so that there is sufficiently old state for 2 latest checkpoints, taking creation time into account as well. I believe `250_000` versions is suitable for the current parameters.

With so little state, creating runtime checkpoint is very fast. If I remember correctly cca 20min and also much less IO intensive than what I believe devops have observed on the Observer nodes (with all historical state)... Consensus checkpoint creation compared to runtime checkpoint creation is instant (seconds).

martintomazic:
We should probably wait with merging this until  https://github.com/oasisprotocol/internal-ops/issues/1221 (snapshots without holes) is ~merged~ finished.

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,78 @@
...
+# ... sections not relevant are omitted ...
+consensus:
+    # ... sections not relevant are omitted ...
+    prune:
+        strategy: "keep_n"
+        num_kept:  <n>
+```
+
+where `<n>` is the number of blocks and state versions that will be kept.
+
```
matevz, line 25: Can you mention what are the defaults currently?

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,78 @@
...
+network health are encouraged to set `n=5_000_000` (approximately 1 year of data).
+
+### Late Pruning
+
+Due to the [LSM Tree][] design of the underlying databases, enabling pruning after
+your node has been running for a while, or changing configuration to retain less data,
+may not work as expected.
+
+To change pruning configuration operators must follow this steps:
+1. Gracefuly shutdown the node (validators should take [special care][graceful shutdown]).
```
matevz, line 43:
```suggestion

1. Gracefuly shutdown the node (validators should take [special care][graceful shutdown]).
```

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,78 @@
...
+2. Configure pruning as described in [Consensus Pruning](#consensus-pruning).
```
matevz:
```suggestion
2. Configure pruning as described in the [Consensus Pruning](#consensus-pruning) section above.
```

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,78 @@
...
+4. Run the [compaction command].
+5. Start the node again.
+
+[LSM tree]: https://en.wikipedia.org/wiki/Log-structured_merge-tree
+[graceful shutdown]: ../maintenance/shutting-down-a-node.md
+[pruning command]: https://github.com/oasisprotocol/oasis-core/blob/master/docs/oasis-node/cli.md#prune-experimental
+[compaction command]: https://github.com/oasisprotocol/oasis-core/blob/master/docs/oasis-node/cli.md#compact-experimental
+
+## Paratime Pruning
+
```
matevz, line 55:
```suggestion
```

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,78 @@
...
+where `<n>` is the number of paratime's state versions that will be kept.
+
+### Suggested Pruning Configuration
+
+Paratime state is much larger than the consensus state. As a consequence, as the
+number of versions in the state database grows, pruning becomes slower and slower.
+For this reason **paratime state pruning must be configured
+from the start or not configured at all**.
+
+To configure pruning it is recommended to set `n=250_000`. The maximum value is
```
matevz:
```suggestion
To configure ParaTime pruning it is recommended to set `n=250_000`. The maximum value is
```

docs/node/run-your-node/advanced/pruning.md
```diff
@@ -0,0 +1,78 @@
...
+`n=400_000`. If you need to preserve more data (e.g. nodes serving historical
+state) you will have to keep the entire state from the genesis.
```
matevz: > you will have to keep the entire state from the genesis

Is this that speceific ParaTime's genesis or consensus? Where does the 400_000 limit come from?

matevz (approved):
I left some suggestions.

Can you also update https://github.com/oasisprotocol/docs/blob/martin/task/update-pruning-documentation/docs/node/run-your-node/prerequisites/hardware-recommendations.md?plain=1#L153-L168
Perhaps just title it "Pruning" and replace the links with the new one to your chapter.

docs/node/run-your-node/advanced/pruning.md
martintomazic: Yes Paratime specific, updated the text to make it clear.

>Where does the 400_000 limit come from

Is not hard limit, but something that as from my experiments am willing to vouch it will still prune with reasonable speed.

docs/node/run-your-node/advanced/pruning.md
martintomazic: Capitalized throughout the file correctly.

docs/node/run-your-node/advanced/pruning.md
martintomazic: Does extra line really make it more readable?

docs/node/run-your-node/advanced/pruning.md
martintomazic: Current defaults are `600` which I find not appropriate as ideally you should have enough data to produce checkpoints.

Might be a good idea to update `oasis-core` defaults to the values suggested here.

martintomazic:
>Perhaps just title it "Pruning" and replace the links with the new one to your chapter.

Added new links, further refinement should be done as part of #1506 as is completely outdated :(.

docs/node/run-your-node/advanced/pruning.md
matevz: It's the markdown standard/linter. Enumerations, headings and code blocks should have empty lines above and below.

martintomazic:
As agreed with the reviewer in private this was ready to be merged. Future refinements can be done in the follow up.

PR #1599, opened 2025-12-08
docs/node/run-your-node: Add compatibility matrix for archive nodes

anzoman:
Documenting compatible versions for setting up Oasis archive nodes for previous network generations.


docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
+---
+description: This page describes how to run an archive node on the Oasis Network.
+---
+
+# Archive Node
+
+This guide will cover setting up an archive node for the Oasis Network. Node
```
peternose:
```suggestion
This guide covers setting up an archive node for the Oasis Network.

An archive node serves historical consensus and runtime state that cannot normally be accessed through other nodes in the network because the state has been pruned or removed during dump-and-restore network upgrades. The node does not actively participate in the consensus protocol and has most of the node features disabled.
```

Let split intro and archive node description.

I would somewhere also mention that the node doesn't need to be registered neither added to an existing entity, but not sure where.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+started in archive mode only serves existing consensus and runtime states.
+The node has all unneeded consensus and P2P functionality disabled, therefore
+it will not participate in the network. Archive nodes can be used to access
+historic state which is pruned in dump-restore network upgrades.
+
+## Prerequisites
```
peternose: Too general, should be avoided.  Maybe change to `historical state` and define below what snapshot is and where can it be accessed.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+Running an archive node requires a pre-existing `oasis-node` state. If you don't have one,
```
peternose: Term `oasis-node state` is confusing. Let's stick with historical consensus and runtime state.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+you can download a snapshot of a specific network state [here][snapshots].
```
peternose: Instead of `here` you should use a proper name, e.g. `on our official Snapshot page` or something better.

Maybe also define what snapshot is, how to extract it and when new snapshots are added.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+[snapshots]: https://snapshots.oasis.io
+
+## Configuration (Oasis Core 23 and later)
```
peternose: The title is too long and intro text is missing. It should be changed to something like:

```
## Configuration (use better name)

Archive nodes can be configured with different historical state, ....
- Oasis Core 22 and earlier
- ...

### Oasis Core 23 and latter

### Oasis Core 22 and earlier
```

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+Starting from the Oasis Core version 23, the configuration for enabling archive mode has changed.
+Use the `mode` setting:
+
+:::info
```
peternose: I would avoid this, as they completely destroy reading flow. Keep it simple.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+:::
+
+## Configuration (Oasis Core 22 and earlier)
+
+For all pre-Eden networks, such as Damask, the configuration remains the same but requires the
+appropriate version of `oasis-node` and the node state.
+
+#### Damask
+
+To run an archive node for Damask, use [Oasis Core v22.2.13] and the following
```
peternose, line 153: How do I know which snapshot to use?

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+    enabled: true
+```
+
+:::warning
+
+Ensure you are using the correct version of oasis-node and the pre-existing state for your specific pre-Eden network.
+
+:::
+
+## Starting the Oasis Node
```
peternose, line 213:
```suggestion
## Starting
```
People might confuse archive and oasis node. Let's have simple sections, e.g. configuration/setup, deployment, ... or configuring archive node, starting/running archive node.

It would be nice to have again intro text for these sections so that people know what is coming, e.g. something like this, but better;

```
To run your own archive node for a specific historical state you need to:
- Download snapshot
- Configure archive node
- Start archive node
- ...

### Download snapshot
### Configure archive node
### Start archive node
```

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+You can start the node by running the following command:
+
+```bash
+oasis-node --config /node/etc/config.yml
+```
+
+### Archive node status
```
peternose: This title can be removed.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+```json
+{
+  // other fields omitted ...
+  "mode": "archive",
+  // ...
+}
+```
+
+## Version Compatibility Matrix
```
peternose: When I read the title I have no idea what this is. So it would be better to rename it to snapshots or historical state, move it higher (because this will be the first thing a user needs to do), describe what snapshot is, and then transition to this compatibility matrix.

peternose (commented):
Nice, very good summary, which probably took a lot of time. I did not check if the matrix is correct and everything works :) 

I posted few comments about how I would organize the guide.

docs/node/run-your-node/archive-node.mdx
peternose: I guess the right order of the guide should be the same as how a user would deploy this:
- Select which historical state you want to serve with archive node
- Download correct snapshot and binary
- Prepare configuration
- Run the node

peternose:
Maybe we should also move seed node and archive node guide lower, as are less important. I would expect the following order: validator, non-validator, compute, client, key manager, archive, seed, sentry.

docs/node/run-your-node/archive-node.mdx
anzoman: Reorganized and reworded as suggested.

docs/node/run-your-node/archive-node.mdx
anzoman: It should be more specific now.

docs/node/run-your-node/archive-node.mdx
anzoman: Agree, updated.

docs/node/run-your-node/archive-node.mdx
anzoman: Sure, updated and added some more details about the snapshots.

docs/node/run-your-node/archive-node.mdx
anzoman: I think that the title "Configuration" is okay, and we also use it everywhere else.
I added the intro text and reworded the titles.

docs/node/run-your-node/archive-node.mdx
anzoman: Removed.

docs/node/run-your-node/archive-node.mdx
anzoman: Now, it's more clear as I reorganized the sections, so you pick the snapshot first.

docs/node/run-your-node/archive-node.mdx
anzoman: We use `Starting the Oasis Node` everywhere, so I think we can stick with that or refactor also other parts of the docs.

Reordering the content and sections makes sense, so I did it.

docs/node/run-your-node/archive-node.mdx
anzoman: Removed.

docs/node/run-your-node/archive-node.mdx
anzoman: Agreed, reordered the sections. Now, it should be more straightforward for the users.

anzoman:
> Maybe we should also move seed node and archive node guide lower, as are less important. I would expect the following order: validator, non-validator, compute, client, key manager, archive, seed, sentry.

Sure, we can do this in the next PRs.

anzoman:
@peternose I've improved the content and reordered the sections as suggested.
Should be easier to follow the docs now.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,242 @@
...
+most of the node features disabled. Archive node does not need to be
+registered neither added to an existing entity.
+
+## Snapshots and Version Compatibility
+
+Running an archive node requires historical consensus and runtime state.
+
+You can download a snapshot of a specific network state from
+the [Official Oasis Network Snapshots page][snapshots].
+The files should be extracted into the right directories within the the
```
peternose:
```suggestion
The files should be extracted into the right directories within the
```

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,242 @@
...
+node datadir.
+
+For example, if your `datadir` is defined as `/node/data`
```
peternose:
```suggestion
For example, if your datadir is defined as `/node/data`
```
I would be consistent and use this in the entire document or nowhere.

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,242 @@
...
+[20210413-20220303]: https://snapshots.oasis.io/#node/testnet/20210413-20220303/
+[20220303-20231012]: https://snapshots.oasis.io/#node/testnet/20220303-20231012/
+[20201001-20201118]: https://snapshots.oasis.io/#node/mainnet/20201001-20201118/
+[20201118-20210428]: https://snapshots.oasis.io/#node/mainnet/20201118-20210428/
+[20210428-20220411]: https://snapshots.oasis.io/#node/mainnet/20210428-20220411/
+[20220411-20231129]: https://snapshots.oasis.io/#node/mainnet/20220411-20231129/
+
+## Prerequisites
+
+Before following this part, make sure you've followed the [Prerequisites](prerequisites) and [Run a Non-validator Node](non-validator-node.mdx) sections and have:
```
peternose:
```suggestion
Before continuing, make sure you've followed the [Prerequisites](prerequisites) and [Run a Non-validator Node](non-validator-node.mdx) sections and have:
```

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,242 @@
...
+
+* Oasis Node binary installed and configured on your system.
+* The chosen top-level `/node` working directory prepared. In addition to `etc` and `data` directories, also prepare the following directories:
+  * `bin`: This will store binaries needed by Oasis Node for running the ParaTimes.
+  * `runtimes`: This will store the ParaTime bundles.
+* Genesis file copied to `/node/etc/genesis.json`.
+* Historical state from the snapshot downloaded and extracted into the right directories within the `/node/data` directory:
+  * `consensus/` for the consensus state.
```
peternose:
```suggestion
  * `consensus` for the consensus state.
```

docs/node/run-your-node/archive-node.mdx
```diff
@@ -0,0 +1,242 @@
...
+            default: info
+genesis:
+    file: /node/etc/genesis.json
+runtime:
+    # Paths to ParaTime bundles for all of the supported ParaTimes.
+    paths:
+        - {{ runtime_orc_path }}
+```
+
+:::info
```
peternose: I would remove info box and keep the text. Maybe even merge it with upper instructions.

peternose (approved):
Looks really good 👍

docs/node/run-your-node/archive-node.mdx
anzoman: Updated.

docs/node/run-your-node/archive-node.mdx
anzoman: Good point, refactored.

docs/node/run-your-node/archive-node.mdx
anzoman: Updated.

docs/node/run-your-node/archive-node.mdx
anzoman: Fixed.

peternose (approved):

PR #1626, opened 2026-01-02
feat: add docs capture and export functionality for llmstxt standard

### LINKED ISSUE #1315: Export LLM/AI version of the docs
Instead of integrating [3rd party LLM](https://github.com/oasisprotocol/docs/pull/964) into our docs let's improve AX and export LLM-understandable material similar to how [ElizaOS docs](https://eliza.how/) does it:
Image

The file then looks like this:
[llms-full.txt](https://github.com/user-attachments/files/21102753/llms-full.txt)

Investigate how to combine all markdown files and host it somewhere inside the docs.

uniyalabhishek:
closes https://github.com/oasisprotocol/docs/issues/1315

preview: https://deploy-preview-1626--oasisprotocol-docs.netlify.app/llms.txt
https://deploy-preview-1626--oasisprotocol-docs.netlify.app/llms-full.txt

### DIFF
--- a/src/plugins/llms/capture.ts
+++ b/src/plugins/llms/capture.ts
@@ -1,0 +1,58 @@
+import * as crypto from 'crypto';
+import * as fs from 'fs';
+import * as path from 'path';
+
+import type {MdastNode} from './types';
+
+export const CAPTURE_TREE_VERSION = 1;
+
+function hashString(value: string): string {
+  return crypto.createHash('sha1').update(value).digest('hex');
+}
+
+export function captureDirForSite(siteDir: string): string {
+  return path.join(siteDir, '.docusaurus', 'llms-export', 'trees');
+}
+
+export function capturedTreePathForSource(
+  sourceAbsPath: string,
+  siteDir: string,
+): string {
+  let absolutePath: string;
+  try {
+    absolutePath = fs.realpathSync(sourceAbsPath);
+  } catch {
+    absolutePath = path.resolve(sourceAbsPath);
+  }
+  return path.join(
+    captureDirForSite(siteDir),
+    `${hashString(absolutePath)}.json`,
+  );
+}
+
+export function wrapCapturedTree(tree: unknown): {
+  version: number;
+  tree: unknown;
+} {
+  return {version: CAPTURE_TREE_VERSION, tree};
+}
+
+export function unwrapCapturedTree(payload: unknown): MdastNode | null {
+  if (!payload || typeof payload !== 'object') return null;
+
+  const record = payload as Record<string, unknown>;
+
+  // Handle versioned format
+  if ('version' in record && 'tree' in record) {
+    const version = Number(record.version);
+    if (version !== CAPTURE_TREE_VERSION) return null;
+    return (record.tree as MdastNode) ?? null;
+  }
+
+  // Legacy format support (version 1 only)
+  if (CAPTURE_TREE_VERSION === 1) {
+    return payload as MdastNode;
+  }
+
+  return null;
+}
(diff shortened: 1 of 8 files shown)


src/plugins/llms/generator.ts
```diff
@@ -0,0 +1,671 @@
...
+      docsByPermalink: params.docsByPermalink,
+      currentPermalink: doc.permalink,
+      siteBase: params.siteBase,
+      baseHeadingLevel: 1,
+      titleToStrip: doc.title,
+      linkTarget: 'md',
+    });
+
+    const slug = doc.permalink.replace(/^\//, '').replace(/\/$/, '') || 'index';
+    const filePath = path.join(params.outDir, `${slug}.md`);
```
matevz, line 464: According to https://llmstxt.org/:

```suggestion
    const filePath = path.join(params.outDir, `${slug}.html.md`);
```

matevz (commented):
First pass. I tested it and works as expected. Benchmarks also didn't show any significant slow downs during compilation.

src/plugins/llms/capture.ts
```diff
@@ -0,0 +1,58 @@
...
+    captureDirForSite(siteDir),
+    `${hashString(absolutePath)}.json`,
+  );
+}
+
+export function wrapCapturedTree(tree: unknown): {
+  version: number;
+  tree: unknown;
+} {
+  return {version: CAPTURE_TREE_VERSION, tree};
```
matevz, line 37: Why do you need to version the temporary generated site map?

src/plugins/llms/generator.ts
```diff
@@ -0,0 +1,671 @@
...
+  if (!doc.source) return null;
+
+  const sourcePath = getSourcePath(doc, siteDir);
+  const cacheKey = `${siteDir}::${sourcePath}`;
+
+  if (treeCache.has(cacheKey)) {
+    return treeCache.get(cacheKey) ?? null;
+  }
+
+  const rawTree = loadCapturedTree(sourcePath, siteDir);
```
matevz, line 80: I'd rather call this siteMap?

src/plugins/llms/generator.ts
uniyalabhishek: I think it's safer to keep it as is. our docs use clean URLs (e.g. /node/grpc), so the mirror is /node/grpc.md. switching to /node/grpc.html.md would be a different URL scheme and isn’t a 1-line swap (we’d need to update URL mapping + rewrite logic)

from llmstxt.org, it says to append .md on the same url:
"We furthermore propose that pages on websites that have information that might be useful for LLMs to read provide a clean markdown version of those pages at the same URL as the original page, but with .md appended. (URLs without file names should append index.html.md instead.)"

also the best examples we have follow the same pattern: https://docs.cdp.coinbase.com/llms.txt, https://docs.elizaos.ai/llms.txt, https://docs.privy.io/llms.txt, https://docs.stripe.com/llms.txt

src/plugins/llms/capture.ts
uniyalabhishek: these files aren’t a sitemap, they’re cached captured [MDAST](https://github.com/syntax-tree/mdast) trees under `.docusaurus/llms-export/trees/*.json` that can persist across cached builds. version is a cheap cache invalidation guard so if we ever change the capture payload/shape, we can bump `CAPTURE_TREE_VERSION` and safely ignore stale cached trees instead of producing incorrect exports

src/plugins/llms/generator.ts
uniyalabhishek: this is just a per build cache of sourcePath -> cleaned captured tree to avoid re-processing, I kept treeCache (or can rename to `docTreeCache` if you prefer that), but siteMap reads like sitemap.xml and can be confusing

matevz (approved):

PR #1673, opened 2026-02-19
docs: add llms usage docs

uniyalabhishek:
Closes https://github.com/oasisprotocol/docs/issues/1665

Preview: https://deploy-preview-1673--oasisprotocol-docs.netlify.app/build/tools/llms

PR #1736, opened 2026-05-21
docs: add docusaurus-plugin-copy-page-button

portdeveloper:
## Summary
- add `docusaurus-plugin-copy-page-button` to the docs dependencies
- register the plugin in `docusaurus.config.ts`

## Why
Oasis docs include protocol, SDK, and developer reference pages that are useful to bring into AI tools. This adds a page-level copy/open button that exports the current page as clean markdown and includes one-click Open in ChatGPT, Claude, and Gemini actions.

I checked the repo for overlapping features before opening this. Oasis already has `llms.txt`, `llms-full.txt`, and Context7 MCP documentation for corpus-level AI ingestion. This PR is different: it adds a page-level UI action for copying or opening the current page directly from the docs page.

The plugin has no runtime dependencies and uses peer dependencies for Docusaurus and React.

## Validation
- `git diff --check`
- package manifest, lockfile, and config sanity check

Not run locally: full docs build, due the dependency install/build footprint in this environment.

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -95,6 +95,7 @@ const config: Config = {
       };
     },
     llmsPlugin,
+    'docusaurus-plugin-copy-page-button',
     [
       '@docusaurus/plugin-client-redirects',
       redirectsOptions,
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@mdx-js/react": "^3.0.0",
     "@mermaid-js/mermaid-cli": "^11.4.0",
     "clsx": "^2.0.0",
+    "docusaurus-plugin-copy-page-button": "^0.6.1",
     "fathom-client": "^3.7.2",
     "linkinator": "^6.0.3",
     "markdownlint-cli2": "^0.11.0",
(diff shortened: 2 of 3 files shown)


matevz (changes requested):
1. This seems like duplicated functionality of llms.txt implemented in https://github.com/oasisprotocol/docs/pull/1626 and https://github.com/oasisprotocol/docs/pull/1673. For every documentation page you can already get its markdown source by simply appending .md at the end of the url. It would be better if we didn't rely on external npm package for this and just wrap the existing markdown code in form of a Copy page button.

2. There is a bug when clicking on a link that includes an anchor, for example https://deploy-preview-1736--oasisprotocol-docs.netlify.app/build/rofl/quickstart#prerequisites
, then the button is not visible.

3. The button does not appear in no-javascript environment (not critical, but would be nice to have at least a link to markdown source).

portdeveloper:
Thanks for the review. Given the existing per-page Markdown support and the preference to avoid an external package for this UI, I’ll close this PR rather than push the package approach further. The anchor-link/no-JS notes are useful feedback; I’ll keep them in mind for the plugin and for any future custom implementation.

PR #1763, opened 2026-06-11
run-your-node: Add instructions for Ubuntu 26.x to TEE prerequisites

matevz:
This PR:
- simplifies TDX installation only covering Ubuntu 26.04 which brings TDX support in the default kernel.
- fixes `mode: client` -> `mode: observer` in ROFL node config.
- Bumps bubblewrap instructions from 22.04 to 24.04 because oasis-node doesn't run anymore on Ubuntu 22.04.

### DIFF
--- a/docs/node/run-your-node/prerequisites/set-up-tee.mdx
+++ b/docs/node/run-your-node/prerequisites/set-up-tee.mdx
@@ -15,7 +15,7 @@ To run SGX/TDX enclaves:
 1. your hardware must have SGX/TDX support,
 2. you must have the latest BIOS updates installed,
 3. you must have SGX/TDX enabled in your BIOS,
-4. you must have the Linux kernel, drivers and software components properly
+4. you must have the Linux kernel, drivers, and software components properly
    installed and running.
 
 [Intel SGX]: https://www.intel.com/content/www/us/en/architecture-and-technology/software-guard-extensions.html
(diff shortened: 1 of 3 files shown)


docs/node/run-your-node/prerequisites/set-up-tee.mdx
```diff
@@ -96,17 +96,18 @@ and look for the following line:
 
 #### Ubuntu 22.04+
 
-A convenient way to install the AESM service on Ubuntu 22.04 systems
-is to use the Intel's [official Intel SGX APT repository](https://download.01.org/intel-sgx/sgx_repo/).
+A convenient way to install the AESM service on Ubuntu 22.04 and newer systems
```
peternose: We could bump this to 24.04, as building latest releases of Oasis node fails on 22.04.

matevz: Building or running?

peternose: Both, I think. One user asked us if we can provide builds for 22.x also, as 24.x probably doesn't work there. And our release process builds it on 24.x. There is a way to built it, one user gave instructions on Discord.

kostko: Yes, let's bump.

kostko (approved):

